PR #80, opened 2013-09-20
FieldTracker can't be inherited from abstract model

rouge8:
I added a failing test case for inheriting a `FieldTracker` from an abstract model. It looks like the `class_prepared` signal isn't sent for abstract classes. :/

### DIFF
--- a/model_utils/tests/models.py
+++ b/model_utils/tests/models.py
@@ -250,6 +250,17 @@ class Tracked(models.Model):
     tracker = FieldTracker()
 
 
+class AbstractTracked(models.Model):
+    name = models.CharField(max_length=20)
+    number = models.IntegerField()
+    mutable = MutableField()
+
+    tracker = FieldTracker()
+
+    class Meta:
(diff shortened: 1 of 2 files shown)


bee-keeper:
Using latest `django-model-utils==3.1.2` and `Django==1.11.11` and it still doesn't seem possible to add fieldtracker to an abstract base class.  Am I missing something or was this really never resolved?

Natim:
@bee-keeper it seems that you are right.
@carljm you seems to have a better understanding of me of what needs to be done there.
Do we want to fix that? Can you provide guidance about how you think would be the best way to tackle this?

@bee-keeper had you had a look at @sbnoemi proposal? Maybe handling this on AbstractModel is something model-util doesn't aim to provide?

Natim:
We are keeping track of the FieldTracker for AbstractModel in #382 
We are blocking on a Django bug for it: https://code.djangoproject.com/ticket/30427#ticket

I am closing this.

PR #95, opened 2013-11-24
Fixed #25 - 3-tuple choices on a StatusModel should use the python identifier.

### LINKED ISSUE #25: Use "Python identifier" from three-tuple Choices in StatusModel
Hello Carl.

I started to use StatusModel this week (amazing tool btw) and with model like that:

```
class MyModel(StatusModel):
    STATUS = Choices(
        ('0', 'nao_atendido', 'Não atendido'),
        ('10', 'em_atendimento', 'Em atendimento'),
        ('11', 'contato_sem_sucesso', 'Contato sem sucesso'),
        ('20', 'atendido_sem_sucesso', 'Atendido sem sucesso'),
        ('21', 'atendido_com_sucesso', 'Atendido com sucesso')
    )
```

If I do:

```
>>> x = MyModel(status=MyModel.STATUS.nao_atendido)
>>> x.status
u'0'
>>> x.0
x.0
  ^
SyntaxError: invalid syntax
```

Same happens with a saved instance.

So, IMO, in case of tree tuple choice, it should use second item instead first item to create managers, and, _never_ use a digit.

I love to write patches for you, but since I'm not needing this feature right now and without time I just created that issue.

kezabelle:
If a 3-tuple Choices variant is used in a StatusModel, rather than use the db representation as the attribute on the Model, use the given python identifier.

Up for discussion:
- I've had to opt for using `isinstance` (or alternatively, `hasattr`) to get the required data out of the Choices instance, as there is no way I can see of treating it like a tuple. It always returns either the display representation or the the db & display representation, depending on what you're doing (iterating, getting via [0], etc). I'm open to suggestions.
- I'm not sure of the backwards compatibility ramifications of essentially pulling a switcheroo on existent code, and what could/should be done about it.

### DIFF
--- a/model_utils/models.py
+++ b/model_utils/models.py
@@ -9,7 +9,7 @@ from django.utils.timezone import now
 from model_utils.managers import QueryManager
 from model_utils.fields import AutoCreatedField, AutoLastModifiedField, \
     StatusField, MonitorField
-
+from model_utils.choices import Choices
 
 class TimeStampedModel(models.Model):
     """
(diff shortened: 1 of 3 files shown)


Natim:
This needs to be rebase to be merged.

PR #133, opened 2014-05-16
Add ability to filter an InheritanceQuerySet by model.

schinckel:
This is based upon the feature in django-polymorphic, where you
can do:

```
SuperClass.objects.instance_of(SubClass)
```

This will result in only objects of the subclass being fetched.

Note: this works with any queryset, keeping the filtering and
ordering applied there.

Is this something that would be considered? I'll write up some tests if that is the case.


Natim:
Superseeded by #387 this seems like a nice addition to me.

Issue #109, opened 2014-02-01
Lying about the modified date for a Timestamped model is difficult.


kezabelle:
A bit of prelude: 
Sometimes it is useful to be able to lie to the database about when things happened. Mostly this kind of thing is useful when copying/cloning existing objects (or their attributes to another instance).
Django doesn't make this easy with it's `auto_now`/`auto_now_add` attributes for date based fields, as they always take precedence:

```
class MyModel(Model):
    created = DateTimeField(auto_now_add=True)
    modified = DateTimeField(auto_now=True)
```

It is impossible to lie to the database efficiently:

```
old = datetime.now() - timedelta(days=365)
obj, created = MyModel.objects.create(created=old, modified=old)
# the inserted object won't have the value held in `old`
obj.created = old
obj.save() 
# now it will.
```

So that rules Django's builtin timestamping out.
In model utils, we have `TimestampedModel` which itself is just a collection of the two Auto fields; using this model we can get slightly further in our lying:

```
class MyModel2(TimestampedModel): pass
```

now we can lie efficiently about our creation time:

```
old = datetime.now() - timedelta(days=365)
obj, created = MyModel2.objects.create(created=old, modified=old)
# the inserted object will have the `old` value for created, but not modified.
```

but not our `modified` time, which doesn't test the model instance for the attr's existance before applying it's own.

Personally, I'd be keen to have the lying be possible, but it would be technically backwards incompatible to do so: users might be supplying a modified date and relying on it _not_ being applied (perhaps by having a blanket set of **kwargs and not popping `modified` out)

As there are no tests I can see demonstrating a preferred intention, I'm opening this for discussion on whether there's merit to perhaps allowing this.

andreoliwa:
I think this issue was fixed in https://github.com/jazzband/django-model-utils/pull/388 with this commit: https://github.com/jazzband/django-model-utils/commit/cfbea2ca89c0262a0576c926d5c525bd3315c6b5

PR #137, opened 2014-07-01
Implemented ability to lie about modified fields

kezabelle:
This would be a first stab at the changes I was describing in #109, to allow for manually setting the `modified` attributes of a `TimeStampedModel` (or, really, the fields backing that model)
- Are there any other Django API methods that need demonstrating?
- Changelog/docs would need updating.

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -30,8 +30,14 @@ class AutoLastModifiedField(AutoCreatedField):
 
     """
     def pre_save(self, model_instance, add):
-        value = now()
-        setattr(model_instance, self.attname, value)
+        if add and hasattr(model_instance, self.attname):
+            # when creating an instance and the modified date is set
+            # don't change the value, assume the developer wants that
+            # control.
+            value = getattr(model_instance, self.attname)
+        else:
+            value = now()
+            setattr(model_instance, self.attname, value)
         return value
 
 
--- a/model_utils/tests/tests.py
+++ b/model_utils/tests/tests.py
@@ -1051,6 +1051,70 @@ class TimeStampedModelTests(TestCase):
         t1.save()
         self.assertTrue(t2.modified < t1.modified)
 
+    def test_overriding_created_via_object_creation(self):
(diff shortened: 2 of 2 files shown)


philippeluickx:
This would be great for testing purposes... anything blocking this PR?

merwok:
I have copied this in my code in order to have fixed values for tests, it works well!

RDIL:
@philippeluickx conflicts.

rjschave:
This PR checks for a primary key to determine if the object had previously been saved.  If there's no primary key, it allows the modified date to be changed by the developer.

#319 checks for a primary key as well.  If there's no primary key, it sets the modified date to the created date.  This was merged to master.

In both cases there is no primary key and there is a modified date (either the default (now) or the overridden value provided by the developer.  I couldn't think of way to distinguish between the two without making other changes.

I added an __init__ method to the AutoLastModifiedField class and set the value to None.  This gives us the ability to check for three scenarios in the pre_save method:

- If no primary key and modified is None:
    - set the modified date to created (address the goal of #319)
- If no primary key and modified has value:
    - developer provided a value and wants to override the date (the goal of this PR)
- If primary key
    - object had previously been saved - just set modified to now

This does change behavior that others may depend on.  That is, modified will not have a default value (now) when the object is constructed.  The value will be set when pre_save is called.  Are there any concerns about this?

If there are no concerns I can put together the code and tests I developed while working on this.  Should I commit to this PR or start a new one (I'm new to contributing to open source)?

auvipy:
you can introduce a new PR and ask me to review that :) definitely, you can take some code/motivation from old pr :)

eraye1:
bump, can we get this in?  :)

auvipy:
Plz rebase

Natim:
Rebased in #388 388

Natim:
Merged with https://github.com/jazzband/django-model-utils/commit/cfbea2ca89c0262a0576c926d5c525bd3315c6b5

PR #147, opened 2014-08-03
Add tests for field tracker methods in save method

treyhunner:
These tests verify that `FieldTracker` methods work as expected inside the `save()` method of a model.

I used a Django signal to inject assertions into a model's `save()` method and I used a context manager to handle the repetitive signal connecting/disconnecting.  Let me know if you think this approach is overkill.

When I cherry-pick 3496fe42916696db36d799facfbf56f9caaa5035 (from #130) onto these changes and run the tests, three of the four tests fail (the .current() function works properly still).

### DIFF
--- a/model_utils/tests/models.py
+++ b/model_utils/tests/models.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 from django.db import models
+from django.dispatch import Signal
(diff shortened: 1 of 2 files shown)


Natim:
I am closing this while we are reworking the FieldTracker feature.

PR #162, opened 2015-03-05
Fixes get_query_set deprecation warning in Django 1.7.

selwin:
Hi there, this should fix a deprecation warning when using `PassThroughManager` in Django 1.7.


romgar:
This MR is no more relevant, as `PassThroughManagerMixin` has been removed in `django-model-utils` `2.4` (from the ChangeLog: `Remove PassThroughManager. Use Django's built-in QuerySet.as_manager() and/or Manager.from_queryset() utilities instead.`)

Issue #164, opened 2015-03-12
Passthroughmanager incompatibility with django-hvad  and django 1.7 


gone:
I have a manager that extends PassthroughManagerMixin and hvad's Translatable manager.

Hvad attempts to grab an attribute as part of loading the manager into memory ala

```
    try:
            obj = getattr(model, relation)
              shared_model = obj.related.model._meta.shared_model
        except AttributeError:
            continue
```

That worked great pre 1.7 - the queryset would load, not find the expected attribute and raise the appropriate error. 1.7 however, this raises an AppCacheNotReady error, which is not caught and our application doesn't start

carljm:
I'm not very enthused about this fix - I understand your situation, but it doesn't really make sense for `PassThroughManager` to blindly translate any `AppCacheNotReady` errors to `AttributeError`. In some cases the attribute ought to actually be there, but the attempt to access it is being made too soon, and `AppCacheNotReady`, not `AttributeError`, would be the right exception to raise.

It seems like a more straightforward fix for your case would be to define a `model` property on your manager that raises `AttributeError`. This will take precedence over the `__getattr__` on `PassThroughManager`.

Or if that doesn't work (maybe because later on the `model` attribute should actually be defined?), you could define your own `__getattr__` in your subclass which wraps a call to `super(...).__getattr__(...)` in the `try: ... except AppCacheNotReady: raise AttributeError` business.

Also 1.7 includes `Manager.from_queryset(...)` built into Django; you may want to consider replacing `PassThroughManager` with that.

gone:
Was unaware that 1.7 had pulled passthrough into core - using that worked great. Thanks for the advice!

PR #165, opened 2015-03-12
fix for 1.7 appcache error attribute error

gone:
Fix for #164


carljm:
Closing; see comment on #164. Thanks for taking the time to put together a PR!

PR #171, opened 2015-06-17
fix deconstruct method in SplitField

pacahon:


carljm:
I don't think this fix is right. I think the current code should be changed, but it should be changed to simply hardcoding `no_excerpt_field=True` in the deconstruct output, since the excerpt field itself will always be deconstructed as well.

pacahon:
Hope it so. Check this idea tomorrow.

PR #173, opened 2015-06-29
Use shields.io for PyPI badge

treyhunner:
pypipins is down: badges/pypipins#37

shields.io seems to be fairly stable.

Some other fun badges:

python
versions
license
stability
issue
count


carljm:
Not sure what's up with Travis, but otherwise this lgtm. And feel free to add any other badges you think would be useful.

treyhunner:
I think I'll stick with this badge for now.

PR #175, opened 2015-06-29
Use codecov instead of coveralls

treyhunner:
The biggest benefit of codecov is that it supports branch coverage whereas Coveralls does not.

Codecov also [has a browser plugin](https://codecov.io/#features) that overlays code coverage on diffs and files.

I do not have admin access, so I believe integrating this will require @carljm or another repo admin to enable the repository [on codecov](https://codecov.io/github/carljm/django-model-utils).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,4 @@ matrix:
     - env: TOXENV=py33-django_trunk
     - env: TOXENV=py34-django_trunk
 
-after_success: codecov
+after_success: codecov --min-coverage=100
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ env:
   - TOXENV=py34-django_trunk
 
 install:
-  - pip install --upgrade pip setuptools tox virtualenv coveralls
+  - pip install --upgrade pip setuptools tox virtualenv codecov
 
 script:
   - tox
@@ -40,4 +40,4 @@ matrix:
     - env: TOXENV=py33-django_trunk
     - env: TOXENV=py34-django_trunk
 
-after_success: coveralls
+after_success: codecov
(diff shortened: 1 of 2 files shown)


treyhunner:
Full disclosure: since opening this, I am now working with/on Codecov

romgar:
This issue on coveralls seems still open, I don't mind to change, specially if it's adding missing features.

I don't have enough rights to move it to codecov though.

carljm:
Wow, this issue is old. Sorry I missed it @treyhunner. I'd be happy to just add both of you as repo admins, but it seems GH doesn't allow me to do that?

I clicked around a bit on codecov but it wasn't clear to me what I need to do to allow access :/

treyhunner:
@carljm: no worries. I haven't used django-model-utils in a while, but I'd be happy to setup codecov when I get the time if you can figure out how to add me as an admin on GitHub. It might not be possible, but I'm not sure. I'm pretty sure Codecov requires a GitHub admin to set it up though.

romgar:
@carljm it seems that we don't even need to configure anything. Following code (displayed on their website) will look for coverage reports and upload the data to codecov:
```
after_success:
  - bash <(curl -s https://codecov.io/bash)
```
Not really sure if you need to give any access before using it...
A specific token is also generated per repository, but we don't need it with Travis-CI (`Repository Upload Token [...] Not required on Travis-CI, CircleCI or AppVeyor for public repositories.`)

romgar:
And it seems that we can only add `collaborators` to a project...

carljm:
Yeah, AFAICT there's no way to add admins to a GitHub project without transferring ownership or moving it to an organization. I don't mind doing it if I can get some more specific guidance on what needs to be done; I clicked around on codecov for a few minutes and couldn't figure it out.

lucaswiman:
Looks like this has been implemented, as codecov is used in the current travis setup.

PR #176, opened 2015-07-01
German translations

Phist0ne:
Includes changes in setup.py to enable automatic compilation and installation of django message catalogs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,6 @@ setup(
     tests_require=["Django>=1.4.2"],
     test_suite='runtests.runtests',
     setup_requires=['Django>=1.4.2'],
-    include_package_data=True,
     package_data = {
         'model_utils': ['locale/*/LC_MESSAGES/django.po','locale/*/LC_MESSAGES/django.mo'],
     },
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,8 @@
 from os.path import join
 from setuptools import setup, find_packages
+from setuptools.command.install_lib import install_lib as _install_lib
+from distutils.command.build import build as _build
+from distutils.cmd import Command
 
 
 long_description = (open('README.rst').read() +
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -40,5 +83,12 @@ def get_version():
     ],
     zip_safe=False,
     tests_require=["Django>=1.4.2"],
-    test_suite='runtests.runtests'
+    test_suite='runtests.runtests',
+    setup_requires=['Django>=1.4.2'],
+    include_package_data=True,
```
carljm: I usually prefer to leave this `False`, since its functionality depends on VCS plugins. The `package_data` kwarg still works without it.

carljm:
This is great, thank you! Any chance you'd be willing to add a paragraph in `CONTRIBUTING.rst` about the steps needed to update translation files if strings change in a pull request?

Phist0ne:
Added a paragraph to CONTRIBUTING.rst and did some minor changes in the po file.
After adding the po files to the MANIFEST.in setup also copies the .mo files to the source distributions if the exist. If this is no issue to you you may pull this as it is.

carljm:
Merged, thank you!

carljm:
Hi @Phist0ne -- it seems like the new stuff in `setup.py` is causing problems for some people installing the new 2.3 release; see #178. It would be best if having the gettext toolset installed were not a requirement for building/installing model-utils; I'm ok with updating the `.po` files manually when strings change.

Do you have any advice on a fix here? If not, I will probably just remove the new code from `setup.py` and make a 2.3.1 release.

Thanks!

pjdelport:
Hi, I just want to add my voice to #178 and #179: the setup.py change also breaks our builds.

Relying on the django-admin management commands or DJANGO_SETTINGS_MODULE in general during setup.py is a bad idea, because there's no way to ensure or know whether the Django environment is fully initialized and configured yet (and it's likely not to be, when this library is still being installed).

carljm:
Yes, @pjdelport is right. I've just pushed a commit to remove all of the translations stuff from `setup.py`, and will release 2.3.1 shortly.

My fault for inadequate review here; I glanced at it briefly and thought it was just adding a new `setup.py` command as a convenience, didn't notice it was being run automatically on build/install.

pjdelport:
@carljm: Thanks!

@Phist0ne: Thanks for the effort taken automate the translation stuff, too: that's always a worthwhile goal, even if these particular changes ended up having unintended side effects.

carljm:
2.3.1 is released. Sorry again for the breakage. And yes, as I said on #181, thank you @Phist0ne for the German translation and the work on automating it. (In 114f4fe2281537 I've now added a `translations.py` script and Makefile tasks as well as docs in `CONTRIBUTING.rst` to make non-automated management of translations hopefully not too hard.)

Issue #178, opened 2015-07-18
Trouble installing 2.3


imkevinxu:
Getting the following error when trying to upgrade to 2.3, any ideas?

```
Running setup.py install for django-model-utils
Complete output from command /app/.heroku/python/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-J22Nrq/django-model-utils/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-473IRE-record/install-record.txt --single-version-externally-managed --compile:
running install
running build
running compile_translations
CommandError: Can't find msgfmt. Make sure you have GNU gettext tools 0.15 or newer installed.

----------------------------------------
Rolling back uninstall of django-model-utils
Command "/app/.heroku/python/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-J22Nrq/django-model-utils/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-473IRE-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-J22Nrq/django-model-utils
```

carljm:
Hrm, the first translation was recently added, and some things were added to `setup.py` that might be causing problems. I'll look into it, thanks for the report.

joeweiss:
Getting the same error here. Thanks.

approxit:
Another problem here.

Requirements installation (pip install -r) is happening before even project files are present, but `ENV`'s are set. Triggering `compilemessages` by `model_utils` can't load project `DJANGO_SETTINGS_MODULE`. Even after reordering a little bit, installing `model_utils` with project files, it fails to load project settings.

Compiling messages on package install is a bad idea. I have not seen such a practice in any other package. Now we know why.

richardasaurus:
On ubuntu, installing `gettext` fixes this

carljm:
I just pushed a commit to fix this, and will issue a 2.3.1 release. Sorry everyone for the hassle of the broken release; my fault for not reviewing that PR closely enough to realize the implications of the `setup.py` changes.

carljm:
2.3.1 is released. Sorry again for the breakage.

Issue #179, opened 2015-07-19
setup_requires causes issues on Python 2.6


moggers87:
If I try to install Django and django-model-utils at the same time, a second copy of Django gets downloaded to satisfy the `setup_requires` line in `setup.py`. This won't normally cause issues, but on Python 2.6 it will produce the following:

```
$ pip install -r test-requires.txt 
Downloading/unpacking Django<1.7 (from -r test-requires.txt (line 1))
  Downloading Django-1.6.11.tar.gz (6.8MB): 6.8MB downloaded
  Running setup.py egg_info for package Django
    warning: no previously-included files matching '__pycache__' found under directory '*'
    warning: no previously-included files matching '*.py[co]' found under directory '*'
Downloading/unpacking django-model-utils (from -r test-requires.txt (line 2))
  Downloading django-model-utils-2.3.tar.gz
  Running setup.py egg_info for package django-model-utils
    Traceback (most recent call last):
      File "<string>", line 16, in <module>
      File "/tmp/pip-build-moggers87/django-model-utils/setup.py", line 92, in <module>
        'compile_translations': compile_translations}
      File "/usr/lib/python2.6/distutils/core.py", line 113, in setup
        _setup_distribution = dist = klass(attrs)
      File "/usr/lib/python2.6/site-packages/setuptools/dist.py", line 221, in __init__
        self.fetch_build_eggs(attrs.pop('setup_requires'))
      File "/usr/lib/python2.6/site-packages/setuptools/dist.py", line 245, in fetch_build_eggs
        parse_requirements(requires), installer=self.fetch_build_egg
      File "/usr/lib/python2.6/site-packages/pkg_resources.py", line 538, in resolve
        dist = best[req.key] = env.best_match(req, self, installer)
      File "/usr/lib/python2.6/site-packages/pkg_resources.py", line 780, in best_match
        return self.obtain(req, installer) # try and download/install
      File "/usr/lib/python2.6/site-packages/pkg_resources.py", line 792, in obtain
        return installer(requirement)
      File "/usr/lib/python2.6/site-packages/setuptools/dist.py", line 293, in fetch_build_egg
        return cmd.easy_install(req)
      File "/usr/lib/python2.6/site-packages/setuptools/command/easy_install.py", line 472, in easy_install
        return self.install_item(spec, dist.location, tmpdir, deps)
      File "/usr/lib/python2.6/site-packages/setuptools/command/easy_install.py", line 502, in install_item
[...]

PR #181, opened 2015-07-20
Fix Issue #178

Phist0ne:
use call_command instead of execute_from_command_line enabling us to handle CommandErrors:


carljm:
Thanks for the PR. More issues have come up over the weekend (see both #178 and #179). I don't think adding `setup_requires` or trying to automatically compile messages on install is a good idea. I'm going to remove all the translation-related changes from `setup.py`.

Thanks again for the contribution of the translations, and for the attempt at automation (and this follow-up fix), even though it doesn't work out in practice.

PR #182, opened 2015-08-14
New PersistentMonitorField type

timmygee:
Hi there. I've found this useful in my code and thought you might want to merge it into the main codebase so it can be used by others

- Same as a MonitorField except once a timestamp is recorded on field change, it is never recorded again for any other change
- Useful if you only ever want the timestamp to be set on the first change

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -126,6 +126,33 @@ class MonitorField(models.DateTimeField):
         return name, path, args, kwargs
 
 
+class PersistentMonitorField(MonitorField):
+    """
+    Like MonitorField, but once the timestamp for the first change is recorded
+    it is never set again
+    """
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('default', now())
+        super(PersistentMonitorField, self).__init__(*args, **kwargs)
+
+    def pre_save(self, model_instance, add):
+        value = now()
+        previous = getattr(model_instance, self.monitor_attname, None)
+        current = self.get_monitored_value(model_instance)
+        if previous != current:
+            monitor_initial = self.get_default()
+            monitor_current = getattr(
+                model_instance, self.attname, monitor_initial)
+            # if this field's timestamp is still the default we record the 
+            # current time value. It will only ever be the default value on
+            # the first save, hence the next timestamp value is persistent
+            if (monitor_current == monitor_initial
+                    and (self.when is None or current in self.when)):
+                setattr(model_instance, self.attname, value)
+                self._save_initial(model_instance.__class__, model_instance)
+        return super(MonitorField, self).pre_save(model_instance, add)
+
+
 SPLIT_MARKER = getattr(settings, 'SPLIT_MARKER', '<!-- split -->')
 
 # the number of paragraphs after which to split if no marker
--- a/model_utils/tests/models.py
+++ b/model_utils/tests/models.py
@@ -7,7 +7,7 @@ from django.utils.translation import ugettext_lazy as _
 from model_utils.models import TimeStampedModel, StatusModel, TimeFramedModel
 from model_utils.tracker import FieldTracker, ModelTracker
 from model_utils.managers import QueryManager, InheritanceManager, PassThroughManager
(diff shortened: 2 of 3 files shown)


model_utils/fields.py
```diff
@@ -126,6 +126,33 @@ def deconstruct(self):
         return name, path, args, kwargs
 
 
+class PersistentMonitorField(MonitorField):
+    """
+    Like MonitorField, but once the timestamp for the first change is recorded
+    it is never set again
+    """
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('default', now())
```
carljm: This isn't right, is it? Sets the default value to the time whenever the model is imported, and won't change until the server restarts.

carljm:
Hi, thanks for the contribution! I have two questions:

1. I've never come across a use case for this variant, and I'm hesitant to keep piling things in without strong evidence that they are useful. Can you enumerate some real-life use cases you've run across? Can you point to any other examples of someone implementing this pattern (showing that the need for it goes beyond just you)?

2. I'm not a fan of this implementation. I think this type of field should be made nullable, and then set on save only if still `None`. The whole default-set-at-import-time-and-then-compared-against doesn't make sense to me, and seems to require much more code (and duplication of the `MonitorField` version of `pre_save` than I think should be necessary to implement this pattern. I think it ought to be implementable with just a single extra conditional in the subclass `pre_save`, to only call the superclass method if the current value of the monitor field is not `None`.

model_utils/fields.py
timmygee: __init__ runs when the model is instantiated, so I don't think that will be an issue. It could be set to None if that's more preferable...

timmygee:
Hi @carljm 

In answer to your points:

> 1. I've never come across a use case for this variant, and I'm hesitant to keep piling things in without strong evidence that they are useful. Can you enumerate some real-life use cases you've run across? Can you point to any other examples of someone implementing this pattern (showing that the need for it goes beyond just you)?

I've been using this in an existing project that manages building inspection and maintenance. When an InspectionItem model instance is created there is a PersistentMonitorField monitoring as to when the InspectionItem moves into any of 3 final statuses. At that point the time needs to be recorded. This time should not ever be overwritten because it's only the first time it occurred that is important since it's a final action. Hence the need for it to be persistent. If a the item needs further inspection a new InspectionItem instance will be created and the same process ensues.

> 2. I'm not a fan of this implementation. I think this type of field should be made nullable, and then set on save only if still None. The whole default-set-at-import-time-and-then-compared-against doesn't make sense to me, and seems to require much more code (and duplication of the MonitorField version of pre_save than I think should be necessary to implement this pattern. I think it ought to be implementable with just a single extra conditional in the subclass pre_save, to only call the superclass method if the current value of the monitor field is not None.

That's fine, I just noticed that the MonitorField will set a value on model instance creation by default. I was therefore just following the same construct for PersistentMonitorField to allow the same flexibility. The way it works currently a timestamp will be set when the model is created (as a default behaviour), just like the way a MonitorField works. Subsequent valid updates will only have an effect the first time the correct criteria is satisfied.

model_utils/fields.py
carljm: I don't think that's right; a model class `__init__` runs when the model is instantiated, but a field `__init__` runs at model import time.

carljm:
Hi @timmygee - after mulling it over a bit more, I'm thinking this is something that's easy enough to do as a subclass that I don't really want to add it to the library, barring evidence that it's something lots of people use.

FWIW, if I did add it the implementation I'd be looking for is a nullable field that stays `None` until the first change of the monitored field. I think that could be implemented with little more than a three or four line override of `pre_save` (though obviously I haven't tried yet). But don't worry about refactoring the implementation at this point; I don't think I'm going to add it.

Thanks again for the offered contribution!

timmygee:
No problem @carljm 

Hopefully I can contribute something more useful at a later stage :)

model_utils/fields.py
timmygee: You're absolutely right. I got confused between field and model declarations. If this were ever to happen a nullable field initialised to None would be the best approach

PR #188, opened 2015-09-29
Updated StatusField to work with django 1.9 #187

### LINKED ISSUE #187: Choices not showing up in django 1.9 alpha1
Following the documentation example in django 1.9a1 gives a status field with empty choices.

That means the following:

```python
>>> MyModel._meta.get_field('status').choices == []
True
```

`MyModel._meta.get_field('status')._choices` it's filled correctly though.

nemesisdesign:
StatusField choices are not set correctly in django 1.9a1.

Would fix #187

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -59,6 +59,7 @@ class StatusField(models.CharField):
                 "To use StatusField, the model '%s' must have a %s choices class attribute." \
                 % (sender.__name__, self.choices_name)
             self._choices = getattr(sender, self.choices_name)
+            self.choices = self._choices  # django 1.9
             if not self.has_default():
                 self.default = tuple(getattr(sender, self.choices_name))[0][0]  # set first as default
 
@@ -68,6 +69,7 @@ class StatusField(models.CharField):
         # the STATUS class attr being available), but we need to set some dummy
         # choices now so the super method will add the get_FOO_display method
         self._choices = [(0, 'dummy')]
+        self.choices = self._choices  # django 1.9
         super(StatusField, self).contribute_to_class(cls, name)
 
     def deconstruct(self):


carljm:
Hi, thanks for the contribution! Unfortunately this seems to break the test suite on all pre-1.9 Django versions (see Travis).

I would be more inclined to merge a PR with fixes for Django 1.9 if it actually fixed _all_ the issues with Django 1.9 and made the test suite pass. There are a number of other problems with 1.9 currently: https://travis-ci.org/carljm/django-model-utils/jobs/71797830

nemesisdesign:
Patch is not backward compatible, I'll have to fix it tomorrow. Any suggestions?

nemesisdesign:
ok @carljm

model_utils/fields.py
```diff
@@ -59,6 +59,7 @@ def prepare_class(self, sender, **kwargs):
                 "To use StatusField, the model '%s' must have a %s choices class attribute." \
                 % (sender.__name__, self.choices_name)
             self._choices = getattr(sender, self.choices_name)
+            self.choices = self._choices  # django 1.9
```
gustavi: try to change this line by : 

```python
try:
    self.choices = self._choices  # django 1.9
except AttributeError:
    pass
````

File: model_utils/fields.py [hunk 2]
```diff
@@ -68,6 +69,7 @@ def contribute_to_class(self, cls, name):
         # the STATUS class attr being available), but we need to set some dummy
         # choices now so the super method will add the get_FOO_display method
         self._choices = [(0, 'dummy')]
+        self.choices = self._choices  # django 1.9
```
gustavi: here too

nemesisdesign:
I haven't had time to keep working on the entire task. I'm not aware if anybody else has solved this specific issue or not. Should I close this or send the suggested changes?

carljm:
The test suite is running on Django 1.9 successfully now, so this is the only 1.9 incompatibility I am aware of. So if the patch is updated to pass on all supported Django versions I will certainly merge it. Thanks!

gustavi:
@carljm could you test the code I propose in my comment ? It should work fine but I have no time to test it at the moment.

carljm:
@gustavi Sorry I'm short on time too, and don't use StatusField anymore myself. I'll be happy to merge a PR that passes all the tests (and adds a new one that fails before this change and passes after it), but I don't have time to do more than that.

nemesisdesign:
This change is not necessary anymore, another patch was merged to fix this issue.

See the current code:

- https://github.com/carljm/django-model-utils/blob/master/model_utils/fields.py#L73-L75
- https://github.com/carljm/django-model-utils/blob/master/model_utils/fields.py#L63-L64

gustavi:
Perfect ! Could you make a realease ?

carljm:
@nemesisdesign Ha, sorry about that! I guess I merged a fix for this in some other PR and didn't even realize it.

@gustavi 2.4 was released a couple weeks ago, and already includes that fix.

nemesisdesign:
@carljm no problem, thanks.

Issue #190, opened 2015-10-27
QuerySet._clone() no longer has 'klass' parameter. [Django 1.9]


jarekwg:
https://github.com/django/django/commit/4c3bfe9053766d378999d06ec34ee5fd4e39f511#diff-5b0dda5eb9a242c15879dc9cd2121379R1043

This breaks PassThroughManager, amongst other things.

carljm:
Thanks for the report. I don't really use `model-utils` anymore (aside from `Choices`), so I'm not likely to find time to update this stuff for Django 1.9. I'll gladly review and merge pull requests, though. (And if anyone wants to take over maintenance of `django-model-utils`, I am more than willing to hand it off.)

jarekwg:
I was just made aware of QuerySet.as_manager(), introduced in Django1.7. It makes PassThroughManager redundant. If you like I can submit a pull request that either
-removes PassThroughManager entirely (dropping support for Django<1.7)
-raises exception when trying to use PassThroughManager for Django>=1.9 (and maybe deprecation warning for Django>=1.7)
-overrides PassThroughManager's constructor to return QuerySet.as_manager() for Django>=1.7

carljm:
Since Django 1.7 is now the minimum version with upstream security support, I'm happy for model-utils master to bump to 3.0alpha and drop support for everything prior to 1.7, and remove PassThroughManager.

jarekwg:
Cool. Made a fork, will make pull request in a coupla days.

PR #191, opened 2015-10-28
Add support for Django 1.9

jarekwg:
In response to #190


model_utils/tests/tests.py
```diff
@@ -868,6 +867,8 @@ def test_select_subclass_just_self(self):
         ])
 
 
+    @skipIf(
```
carljm: Hmm, first thing in this PR I'm not quite happy with. Did you try adjusting this test to use some other unrelated model from this app's test models, rather than using `auth.User`?

tox.ini
```diff
@@ -2,8 +2,8 @@
 envlist =
     py26-django{14,15,16},
     py27-django14, py27-django15_nosouth,
-    py{27,32,33}-django{15,16,17,18,_trunk},
-    py34-django{17,18,_trunk},
+    py{27,32,33}-django{15,16,17,18,19_trunk},
```
carljm: missing a comma here, should be `...,19,_trunk}`

tox.ini
```diff
@@ -2,8 +2,8 @@
...
+    py34-django{17,18,19_trunk},
```
carljm: missing comma here too

carljm:
Hmm, Travis build is failing, but it looks like an incompatibility between Django 1.9b1 and Python 3 -- not sure how that made it past Django's CI. Guess we could disable CI for Django 1.9 / Python 3 for now? Don't want to merge a red build.

Thanks for putting this together!

jarekwg:
Nah i'll figure out what's wrong with python3 compatibility. Cheers for the fast feedback.

model_utils/tests/tests.py
jarekwg: Fair point. Will do.

jarekwg:
Alright I think that's everything covered.

carljm:
Looks great, thanks!

PR #194, opened 2015-11-12
Fixes AttributeError - ModelChoiceField requires _prefetch_related_lookups attribute

bogdal:
This PR fixes the following error in the latest stable version of Django (`1.8.6`):
```
(...)/site-packages/django/forms/models.py", line 1095, in __iter__
    method = 'all' if self.queryset._prefetch_related_lookups else 'iterator'
AttributeError: 'InheritanceManager' object has no attribute '_prefetch_related_lookups'
```
`ModelChoiceField` requires  [_prefetch_related_lookups](https://github.com/django/django/pull/5397/files#diff-70af885c2725fe87eb3b99a393268d10R1106) queryset attribute.

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -175,6 +175,7 @@ class InheritanceQuerySetMixin(object):
 
 
 class InheritanceManagerMixin(object):
+    _prefetch_related_lookups = []
     use_for_related_fields = True
 
     def get_queryset(self):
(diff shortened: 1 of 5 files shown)


kezabelle:
This issue will presumably be fixed in 1.8.7, as [it's a regression](https://code.djangoproject.com/ticket/25683) which has been backported to `1.9.x` and `1.8.x`. I don't know what the expected timeframe is on the next patch release for Django to know whether it's worth fixing here vs. waiting it out.

carljm:
Thanks for the patch, but I would prefer to wait for the fix in Django. I don't think it will be more than a week or so before 1.8.7 is released.

PR #197, opened 2015-12-02
Start testing against Django 1.9 stable

patrys:
A release that supports 1.9 would be nice.


carljm:
Thanks!

carljm:
2.4 is released.

Issue #184, opened 2015-08-24
get_all_related_objects deprecation warning with Django 1.8


pmdarrow:
https://github.com/carljm/django-model-utils/blob/master/model_utils/managers.py#L103 raises the following deprecation warning in Django 1.8: 

```django.utils.deprecation.RemovedInDjango110Warning: 'get_all_related_objects is an unofficial API that has been deprecated. You may be able to replace it with 'get_fields()'```

hellvix:
I got the same warning on 1.9.

PR #202, opened 2016-01-06
Fix issue #184: deprecation warning in Django 1.8/1.9

boltomli:


carljm:
Thanks for the PR. I'm afraid that we need to fix this in a way that maintains compatibility with Django versions prior to 1.8 (see the Travis-CI results -- tests are failing on all previous Django versions), which means we can't just switch to `get_fields` without a fallback. We need either a version check or a `getattr` based fallback.

boltomli:
Yes you're right. Travis-CI is good to catch it. I'll use it for my 1.9 only test for now.

PR #204, opened 2016-01-12
Django bug has been fixed

mjtamlyn:
The Django bug mentioned by this comment was fixed 3 years ago. I haven't double checked that InheritanceManager now behaves correctly in that use case, but the note at least is invalid.

### DIFF
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -84,14 +84,8 @@ If you don't explicitly call ``select_subclasses()`` or ``get_subclass()``,
 an ``InheritanceManager`` behaves identically to a normal ``Manager``; so
 it's safe to use as your default manager for the model.
 
-.. note::
-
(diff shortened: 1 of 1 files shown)


carljm:
I'm pretty sure that note was added after the bug had already been fixed; observe that the note specifically says "in Django versions prior to 1.6," which are the Django versions where the bug is not fixed. Model-utils still supports all the way back to 1.4, so the note remains relevant.

Probably we should drop support for unsupported Django versions in the next release; at that point the note can be removed. I'll leave this open as a reminder of that.

Thanks for the PR!

mjtamlyn:
Haha, I should read properly.

romgar:
Thanks @mjtamlyn for the PR! The note has been removed in #212, with the unsupported Django versions drop.

PR #208, opened 2016-02-08
Add CI on Python 3.5, drop Python 3.2.

carljm:


bogdal:
Django `1.9+` [doesn't support Python `3.3`](https://docs.djangoproject.com/en/dev/releases/1.10/). I think we should remove `py33-django_trunk` tox env.

carljm:
Yep, force-pushed an updated commit to fix that. Looks all good now! Merging.

PR #211, opened 2016-02-19
FieldTracker extension to emit custom signals on model instance changes.

sandeepklr:
https://github.com/carljm/django-model-utils/issues/210

### DIFF
--- a/model_utils/__init__.py
+++ b/model_utils/__init__.py
@@ -1,4 +1,4 @@
 from .choices import Choices
-from .tracker import FieldTracker, ModelTracker
+from .tracker import FieldTracker, ModelTracker, FieldTrackerSignal
 
 __version__ = '2.4.1a1'
--- a/model_utils/signals.py
+++ b/model_utils/signals.py
@@ -1,0 +1,4 @@
+import django.dispatch
+
+field_tracker_presave = django.dispatch.Signal(providing_args=["sender", "instance", "changed"])
+field_tracker_postsave = django.dispatch.Signal(providing_args=["sender", "instance", "created", "changed"])
--- a/model_utils/tests/models.py
+++ b/model_utils/tests/models.py
@@ -5,7 +5,7 @@ from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 
 from model_utils.models import TimeStampedModel, StatusModel, TimeFramedModel
-from model_utils.tracker import FieldTracker, ModelTracker
+from model_utils.tracker import FieldTracker, ModelTracker, FieldTrackerSignal
 from model_utils.managers import QueryManager, InheritanceManager
 from model_utils.fields import SplitField, MonitorField, StatusField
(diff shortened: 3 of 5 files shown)


sandeepklr:
@carljm Haven't edited AUTHORS and CHANGES yet, will do after PR is reviewed and changes accepted.

carljm:
@sandeepklr Can you clarify what (if anything) is now possible with this feature that isn't already possible by simply listening to Django's built-in pre-save or post-save signals (which send the model object along as an argument) and then checking the information stored in that model instance's field-tracker? Why is a new pair of signals needed?

sandeepklr:
@carljm Instead of including pre-save and post-save logic to "detect" and hold onto differences between pre-save and post-save in every model class, this allows a client to simply register to receive those changes either before a save to the DB or after. Now, a client can simply register to receive the field_tracker_postsave for a particular model and can be made aware of changes. For me personally, this is a way to help encapsulate repeated logic of checking for model instance differences into a custom signal.

carljm:
@sandeepklr Can you give an example of the type of repetitive code these custom signals can replace?

I don't understand why you would need to "detect and hold onto" changes if using the built-in signals, if the model has a `FieldTracker` on it which is already doing that for you.

sandeepklr:
@carljm Apologies for the tardy response. 

This is my use case: I want to start a background task after a model instance has been saved(let's call it the current save) based on changes that occurred to this model instance between the last save() and this current one. At the moment, a FieldTracker only returns in-memory changes to the model instance since the last save(). Once save() is called again to save the changes, FieldTracker starts anew.

So, in order to hold onto changes post save, this is the pattern I would use:


```
class Foo(Model):

	field1 = 
	field2 = 
	
	tracker = FieldTracker(...)
	
	@receiver(presave, sender=Foo)
	def receiver_presave(instance, **kwargs):
		if instance.tracker.changed():
			instance.__prev = instance.tracker.changed()
			
	@ receiver(postsave, sender=Foo)
	def receiver_postsave(instance, **kwargs):
		if instance.__prev:
			start_background_task(instance.__prev)
		
		del instance.__prev
	
```

I find that in my code I have to include this pattern in every Model class that I want to track for changes for my use-case. With the signals, this reduces to the following on the client side:

```
@receiver(field_tracker_postsave, sender=Foo)
def fieldtracker_postsave(instance, changed, created):
       start_background_task(changed)
```

carljm:
Aha, I understand the problem now. I'm still not convinced that additional signals are the right solution to that problem, though -- these added signals just seem too closely duplicative of the existing pre-save and post-save signals. ISTM that a better solution would be to modify `FieldTracker` so that it stores multiple "layers" of changes, rather than just one layer. That way, you could still listen to the regular post-save signal, but get access to the changes from that save.

Natim:
We are keeping track of the FieldTracker for AbstractModel in #382 
We are blocking on a Django bug for it: https://code.djangoproject.com/ticket/30427#ticket

I am closing this.

PR #212, opened 2016-03-27
Drop unsupported Django versions

### LINKED ISSUE #183: Remove Django itself from install_requires
There's an unfortunate side effect of referencing Django itself in setup.py or requirements.txt

It rather unexpectedly results in an upgrade to Django itself if I type: "pip install django-model-utils -U"

And really - your app doesn't 'require' django as much as it's an add-on for Django so not much is lost by removing it. Is anyone likely to install django-model-utils and expect it to install Django for them? 

There's a discussion on the same issue here: https://github.com/charettes/django-admin-enhancer/issues/23

rpkilby:
- Drops support for anything below Django 1.8 (current LTS release)
- ~~Cleans up some pep8 E303 warnings. I can remove this commit - just went ahead because my editor lit up like a christmas tree.~~
- ~~Resolves 183~~.
- Supersedes / closes #162.
- Supersedes / closes #204.

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -5,12 +5,8 @@ from django.db.models.fields.related import OneToOneField, OneToOneRel
 from django.db.models.query import QuerySet
 from django.core.exceptions import ObjectDoesNotExist
 
-try:
-    from django.db.models.constants import LOOKUP_SEP
-    from django.utils.six import string_types
-except ImportError:  # Django < 1.5
-    from django.db.models.sql.constants import LOOKUP_SEP
-    string_types = (basestring,)
+from django.db.models.constants import LOOKUP_SEP
+from django.utils.six import string_types
 
 
 class InheritanceQuerySetMixin(object):
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -139,10 +139,10 @@ class InheritanceQuerySetMixin(object):
         if levels:
             levels -= 1
         while parent_link is not None:
-            if django.VERSION < (1, 8):
-                related = parent_link.related
-            else:
+            if django.VERSION < (1, 9):
                 related = parent_link.rel
+            else:
+                related = parent_link.remote_field
             ancestry.insert(0, related.get_accessor_name())
             if levels or levels is None:
                 if django.VERSION < (1, 8):
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,30 +3,18 @@ language: python
 python: 2.7
 
 env:
-  - TOXENV=py26-django14
-  - TOXENV=py26-django15
-  - TOXENV=py26-django16
-  - TOXENV=py27-django110
-  - TOXENV=py27-django14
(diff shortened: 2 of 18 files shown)


romgar:
I'm not really convinced that we should drop support for old versions if it's not a really big issue.

I agree, in theory we will have a cleaner code, but that will prevent people that can't afford to upgrade their project to use these libraries.

rpkilby:
> but that will prevent people that can't afford to upgrade their project to use these libraries.

There's nothing preventing them from using older versions of the package.

romgar:
I get your point @rpkilby. 

Usually, I'm definitely dropping compatibility when I need to develop new functionalities that would be really annoying/impossible to implement in old versions with compatible code, not only for the sake of code cleaning and following Django support cycle.

@carljm what was your compatibility policy on this package until now, and your views about it?

romgar:
After extra thoughts, I think we shouldn't encourage people to stay on old unsupported/insecure versions of Django by making the third-party apps compatible with these old versions.

@rpkilby would you mind updating your PR?

carljm:
My policy up til now has been "drop support for Django versions when it becomes inconvienent to support them" :-) If there is more active maintenance now, I have no problem with a policy to drop support for Django versions without upstream support.

romgar:
@rpkilby would you mind updating your PR to resolve conflicts? It would be nice to have it live!

rpkilby:
Sure - it completely slipped my mind

rpkilby:
The PR has been updated.

romgar:
Great @rpkilby! I will have a look really soon!

model_utils/managers.py
```diff
@@ -143,10 +139,10 @@ def _get_ancestors_path(self, model, levels=None):
         if levels:
             levels -= 1
         while parent_link is not None:
-            if django.VERSION < (1, 8):
-                related = parent_link.related
-            else:
+            if django.VERSION < (1, 9):
                 related = parent_link.rel
```
romgar: I guess this line is generating a warning in Django `1.10`?

romgar:
You did even a bit more than dropping unsupported versions, but also fixed some deprecation warnings!

model_utils/managers.py
rpkilby: Yep https://docs.djangoproject.com/en/1.10/releases/1.9/#field-rel-changes

romgar:
@carljm I will merge this MR soon. It might be a good idea to do an official release with the latest version before this merge, as some fixes are already in master and people still on old versions could benefit from them.

carljm:
@romgar 2.6.1 is now on PyPI, thanks for the nudge.

romgar:
@rpkilby do you mind resolving the conflict with newly updated `CHANGES` file? I merge it right after.

rpkilby:
Done - the changes have been added to the unreleased section.

romgar:
Great job @rpkilby, thanks a lot!

PR #219, opened 2016-05-25
Fixed a bug with Django >= 1.9 and values_list being called on InheritanceQuerySet

artisavotins:
After upgrading an application to Django 1.9 that uses django-model-utils for polymorphic models, I encountered a bug:

Whenever `values_list` was called on an InheritanceQuerySet, `AttributeError` would be raised complaining about either `tuple`(in case of `flat` parameter being false) or `integer`(or any other primitive variable type that's defined for the model and used in `values_list`) in case of `flat` being true, not having an attribute of `insert_child_model_name_here`.

This is a bugfix for that, which adds type checking to `_get_sub_obj_recurse` function.

I also added a new test case `test_dj19_values_list_on_select_subclasses`, that exercises this bug.

Fastest way to reproduce:

- Check out this branch
- Comment out my changes in managers.py L165:168
- Run test cases

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -2,6 +2,7 @@ ad-m <github.com/ad-m>
 Alejandro Varas <alej0varas@gmail.com>
 Alex Orange <crazycasta@gmail.com>
 Andy Freeland <andy@andyfreeland.net>
+Artis Avotins <artis.avotins@gmail.com>
 Bram Boogaard <b.boogaard@auto-interactive.nl>
 Carl Meyer <carl@dirtcircle.com>
 Curtis Maloney <curtis@tinbrain.net>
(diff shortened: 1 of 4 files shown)


carljm:
Looks great, thank you! Would you mind also updating the changelog with a note about this change? And add yourself to the Authors file, if you feel so inclined?

carljm:
(Please comment once that's done; I don't get notifications when new commits are pushed.)

artisavotins:
I added an entry in the change log inside _master (unreleased)_ paragraph. I also added myself into the Authors file, thanks for that. That will be motivation to commit more often.

I squashed all my commits into one to keep commit history clean.

carljm:
Excellent, thank you! Merging.

PR #233, opened 2016-08-15
Django 1.10 support for FieldTracker

### LINKED ISSUE #232: FieldTracker failing with Django 1.10
I know you're already aware the dj110 tests don't pass, but thought I'd leave an open ticket for this, with (hopefully relevant) info I've come across.

I might also have a go at tackling this over the weekend, but no promises. ;)

So my issue since upgrading to dj110 and updating to model-utils 2.5.1:
Creating a model instance via a `DjangoModelFactory` (in my case using factory-boy), causes the initialised fields to come up on the `FieldTracker` as having changed in successive saves. `DjangoModelFactory()` calls the model's `.save()` method up to three times in some cases, so with this bug, a post_save handler will think the field has changed numerous times and will register fields as having changed multiple times, instead of not mentioning anything since the fields were set on init. This appears to happen both with FK fields and regular ones, so no discrimination there. factory-boy hasn't changed recently, so I suspect the problem lies with model-utils' `FieldTracker`, but I might be wrong.

jarekwg:
Fixes #232 

So two things going on here.

First, as of Django 1.8, we should be using `get_deferred_fields` rather than manually probing for `isinstance(field_obj, DeferredAttribute)`.

Second, as of Django 1.10, it seems inherited models don't pull through the deferred fields in `__class__.__dict__`. Have to consult the base class to get them. I have a feeling this might be a regression and will probs raise a ticket with django to make sure, but this bandaid should work for the time being. There's a slight niggle in that dj110 now supports overriding fields from the base class. I _think_ I'm building the dict in the correct order so that stuff is overridden correctly, but there might need to be additional tests added to check for this (can open a new issue). Also, perhaps there's a way to get a hold of the deferred field that avoids the use of `__class__.__dict__` altogether?

Tests all pass now so removed dj110 from allowed failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,13 +35,4 @@ install:
 script:
   - tox
 
-matrix:
-  allow_failures:
-    - env: TOXENV=py27-django110
-    - env: TOXENV=py34-django110
-    - env: TOXENV=py35-django110
-    - env: TOXENV=py27-django_trunk
-    - env: TOXENV=py34-django_trunk
-    - env: TOXENV=py35-django_trunk
-
 after_success: coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,4 +35,10 @@ install:
 script:
   - tox
 
+matrix:
+  allow_failures:
+    - env: TOXENV=py27-django_trunk
+    - env: TOXENV=py34-django_trunk
+    - env: TOXENV=py35-django_trunk
+
 after_success: coveralls
--- a/model_utils/tracker.py
+++ b/model_utils/tracker.py
@@ -69,6 +69,8 @@ class FieldInstanceTracker(object):
 
         class DeferredAttributeTracker(DeferredAttribute):
             def __get__(field, instance, owner):
+                if instance is None:
+                    return field
                 data = instance.__dict__
                 if data.get(field.field_name, field) is field:
                     instance._deferred_fields.remove(field.field_name)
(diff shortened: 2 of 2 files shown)


jarekwg:
After some [django digging](https://code.djangoproject.com/ticket/27065), a handle to `DeferredAttribute` is now obtainable by accessing the field through the instance's class. I've updated the PR to use this implementation, as it's considerably neater.

.travis.yml
```diff
@@ -35,13 +35,4 @@ install:
 script:
   - tox
 
-matrix:
-  allow_failures:
-    - env: TOXENV=py27-django110
-    - env: TOXENV=py34-django110
-    - env: TOXENV=py35-django110
-    - env: TOXENV=py27-django_trunk
```
carljm: I'd rather leave the trunk builds in `allow_failures` even if they pass currently. Once we get into the heat of 1.11 feature development, they will probably break again, and I don't want the build breaking due to changes outside model-utils' control. Django trunk build is just advance warning, it's not a blocker.

jarekwg: Oh `True`, will do!

model_utils/tracker.py
```diff
@@ -76,19 +79,31 @@ def __get__(field, instance, owner):
...
-                # Django 1.4
-                model = None
-                if hasattr(field_obj, 'model_ref'):
-                    model = field_obj.model_ref()
-
+        if django.VERSION >= (1, 8):
+            self.instance._deferred_fields = self.instance.get_deferred_fields()
+            for field in self.instance._deferred_fields:
+                if django.VERSION >= (1, 10):
+                    field_obj = getattr(self.instance.__class__, field)
```
carljm: Does this version really not work in versions prior to 1.10? It looks to me like it should...

jarekwg: It doesn't.
[This change](https://github.com/django/django/commit/7f51876f99851fdc3fef63aecdfbcffa199c26b9#diff-1e7fc0d7d1b36358e371fab97bd1ddb1R99) is what enabled it.

carljm: Ah, got it.

PR #236, opened 2016-09-12
Add SoftDeletableModel

radowit:
SoftDeletableModel and friends - abstract model that instead of deleting entry marks it as deleted.


carljm:
Looks great, thanks!

PR #238, opened 2016-10-14
Remove Django itself from install_requires

### LINKED ISSUE #183: Remove Django itself from install_requires

andybak:
Fixes #183

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,6 @@ setup(
     author_email='carl@oddbird.net',
     url='https://github.com/carljm/django-model-utils/',
     packages=find_packages(),
-    install_requires=['Django>=1.4.2'],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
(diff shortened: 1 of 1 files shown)


carljm:
Can you also add a note to the changelog about this change, and explaining the reasoning?

(I assume you've verified that the problematic behavior of `pip -U` hasn't changed in recent pip versions?)

romgar:
Following discussion on #183, closing it for now.

PR #239, opened 2016-11-09
SoftDeletableModel: use correct DB connection

browniebroke:
When deleting a SoftDeletableModel instance, the `using` parameter should be passed down to the `save()` method.

https://docs.djangoproject.com/en/1.10/topics/db/multi-db/#selecting-a-database-for-save


carljm:
Thanks!

PR #240, opened 2016-11-16
Ability to purge from db a SoftDeletableModel

browniebroke:
This is useful when testing and in some edge cases.


carljm:
Shouldn't need to copy the method; can use `super()` to call the parent `delete()` method.

browniebroke:
Sorry, wasn't thinking straight, I will update soon.

browniebroke:
Updated to call the parent `delete()` and squashed commits.

charettes:
What about adding a `soft=True` kwarg to `delete()` instead?

carljm:
Yeah, I'd be fine with an added kwarg to `delete`, tbh.

carljm:
LGTM, merging. Thanks for making the updates!

Issue #241, opened 2016-11-18
Recursion error with double MonitorField and defer()/only() usage


romgar:
Using two `MonitorField` in a same model generates a recursion error as soon as we use `only()` or `defer()` on that model, but only with Django `1.10+`.

I will create a PR to demonstrate the issue.

carljm:
Thanks @romgar for the report and failing tests! Unlikely that I will have time to work on a fix for this soon, but I'd be happy to review and merge a fix if someone comes up with one.

romgar:
Issue fixed by PR #242!

PR #242, opened 2016-11-18
Tests to demonstrate recursion error with double MonitorField and defer()/only() usage

romgar:
This PR add tests to explain the issue #241.

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -110,8 +110,11 @@ class MonitorField(models.DateTimeField):
         return getattr(instance, self.monitor)
 
     def _save_initial(self, sender, instance, **kwargs):
+        if self.monitor in instance.get_deferred_fields():
+            # Fix related to issue #241 to avoid recursive error on double monitor fields
+            return
         setattr(instance, self.monitor_attname,
-                self.get_monitored_value(instance))
+            self.get_monitored_value(instance))
 
     def pre_save(self, model_instance, add):
         value = now()
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -110,7 +110,7 @@ class MonitorField(models.DateTimeField):
         return getattr(instance, self.monitor)
 
     def _save_initial(self, sender, instance, **kwargs):
-        if self.monitor in instance.get_deferred_fields():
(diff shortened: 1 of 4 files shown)


carljm:
Tests here are enough to verify that queries don't break, but they don't check whether `MonitorField` actually works in this scenario, if the deferred field is later updated. Looks to me like this fix doesn't handle that case?

romgar:
@carljm do you mean something like that:

```
from freezegun import freeze_time

def new_test(self):
    obj = DoubleMonitored.objects.defer('name').get(name='Charlie')

    with freeze_time("2016-01-14"):
        obj.name = 'Charlie2'
        obj.save()
    self.assertEqual(obj.name_changed, datetime(2016, 1, 14))
```

carljm:
Right, that looks good. Does that test pass with the current fix?

romgar:
Yup! I willl update the PR!

romgar:
@carljm updated!

Any reason to keep all the tests in only one file? And to include them in the python package install?

carljm:
I see, I guess it still works because deferred field loading uses `refresh_from_db`, which loads a new instance from the db, triggering `post_init` signal.

Why is this issue only triggered by having two `MonitorFields` on the same model? Can you confirm that is actually necessary to repro? The fix doesn't seem specific to multiple MonitorFields.

romgar:
I'm not yet crystal clear why it happens with **two** `MonitorField`, I need to investigate further.

I have already tested with only one `MonitorField`, and the recursion didn't happen (even without this patch).

romgar:
I have the feeling that's related to the `post_init` signal registered twice if we have 2 `MonitorField` in the same model.

carljm:
Ok, that makes sense. I think what happens without this fix is that the two MonitorFields bounce back and forth, because checking the one field triggers it loading from db, which sends post_init on an instance missing the second monitored field, which triggers loading the second field, which sends post_init on another new instance that's missing the first field, and they just go back and forth forever.

It seems to me that there's some strange unintended behavior around the new `refresh_from_db` and the `post_init` signal -- I'm not sure `post_init` should really be sent on every `refresh_from_db`. So I think there may be a bug in Django lurking here. But since 1.10 is already released, I'm ok with this workaround for now, even if we get the bug fixed for 1.11.

PR #243, opened 2016-11-20
Update tests to use freezegun to avoid time resolution issues on Windows

### LINKED ISSUE #206: [tests] timestamp resolution is too low on Windows, causes tests to fail
sqlite3
django.1.8

https://github.com/carljm/django-model-utils/blob/master/model_utils/tests/tests.py#L163

Datetime is equal. Resolution of time is too low, for example: 2016-01-28 13:13:27.675000+00:00.

There could be greater than or equal to?!

romgar:
Solution proposed to fix #206 

On a general way, we should use `freezegun` if we want to test anything related to values generated with `now()`.

### DIFF
--- a/model_utils/tests/tests.py
+++ b/model_utils/tests/tests.py
@@ -7,6 +7,8 @@ try:
 except ImportError: # Python 2.6
     from django.utils.unittest import skipUnless
 
+from freezegun import freeze_time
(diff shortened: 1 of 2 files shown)


carljm:
Thanks, merged!

PR #250, opened 2017-01-05
Fix issue when extend QuerySet and Manager - fixes #249

### LINKED ISSUE #249: Issue when trying to add methods to SoftDeletableQuerySet/Manager
I tried to extend `SoftDeletableQuerySet` and `SoftDeletableManager` by subclassing them and adding some method to them in my project, and I've done something like that:

```python
class PostQuerySet(SoftDeletableQuerySet):

    def only_read(self):
        return self.filter(is_read=True)


class PostManager(SoftDeletableManager):
    _queryset_class = PostQuerySet

    def only_read(self):
        return self.get_queryset().only_read()


class Post(SoftDeletableModel):
    is_read = models.BooleanField(default=False)

    objects = PostManager()
```

And then running:
```python
PostManager.objects.only_read()
```

I get: `AttributeError: 'SoftDeletableQuerySet' object has no attribute 'only_read'`.

It's because the manager's `get_queryset` method references `SoftDeletableQuerySet` in `get_queryset` rather than using the `self._queryset_class`. A few other Managers are doing the same, could fix them all at once.

Also, as a side note, we cannot benefit from the [`QuerySet.as_manager()`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#as-manager) because it's new in Django 1.7 and models-utils supports previous versions until #212 is merged.

browniebroke:
Fixes issue described in #249. Add tests to cover the use case.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,9 +5,10 @@ master (unreleased)
 -------------------
 
 * Fix infinite recursion with multiple `MonitorField` and `defer()` or `only()`
-  on Django 1.10+. Thanks Romain Garrigues. Merge of #242, fixes #241.
-* Fix issue #249 when extending `Manager` and `QuerySet` classes to add custom
-  methods.
+  on Django 1.10+. Thanks Romain Garrigues. Merge of GH-242, fixes GH-241.
+* Fix `InheritanceManager` and `SoftDeletableManager` to respect `
+  `self._queryset_class` instead of hardcoding the queryset class. Merge of
+  GH-250, fixes GH-249.
 * Add mixins for `SoftDeletableQuerySet` and `SoftDeletableManager`, as stated
   in the the documentation.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ master (unreleased)
 
 * Fix infinite recursion with multiple `MonitorField` and `defer()` or `only()`
   on Django 1.10+. Thanks Romain Garrigues. Merge of GH-242, fixes GH-241.
-* Fix `InheritanceManager` and `SoftDeletableManager` to respect `
+* Fix `InheritanceManager` and `SoftDeletableManager` to respect
   `self._queryset_class` instead of hardcoding the queryset class. Merge of
   GH-250, fixes GH-249.
 * Add mixins for `SoftDeletableQuerySet` and `SoftDeletableManager`, as stated
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,11 @@ master (unreleased)
 -------------------
 
 * Fix infinite recursion with multiple `MonitorField` and `defer()` or `only()`
-on Django 1.10+. Thanks Romain Garrigues. Merge of #242, fixes #241.
+  on Django 1.10+. Thanks Romain Garrigues. Merge of #242, fixes #241.
+* Fix issue #249 when extending `Manager` and `QuerySet` classes to add custom
+  methods.
+* Add mixins for `SoftDeletableQuerySet` and `SoftDeletableManager`, as stated
+  in the the documentation.
 
 
 2.6 (2016.09.19)
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -192,11 +192,16 @@ class InheritanceQuerySetMixin(object):
         return levels
 
 
+class InheritanceQuerySet(InheritanceQuerySetMixin, QuerySet):
+    pass
+
+
 class InheritanceManagerMixin(object):
     use_for_related_fields = True
+    _queryset_class = InheritanceQuerySet
 
     def get_queryset(self):
-        return InheritanceQuerySet(self.model)
+        return self._queryset_class(self.model)
 
     get_query_set = get_queryset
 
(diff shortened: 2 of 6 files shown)


model_utils/managers.py
```diff
@@ -260,11 +261,16 @@ def delete(self):
...
+class SoftDeletableQuerySet(SoftDeletableQuerySetMixin, QuerySet):
+    pass
+
+
+class SoftDeletableManagerMixin(object):
     """
     Manager that limits the queryset by default to show only not removed
     instances of model.
     """
+    use_for_related_fields = True
```
carljm: This looks like an unrelated change that doesn't belong in this PR. It may be a warranted change, but it should have its own issue and PR and be discussed on its own merits. AFAICT it isn't mentioned in #249 or this PR description.

browniebroke: Correct, but since all the others Managers in that file have it, I've added it as it looked consistent with the project (also probably because this is what I'm doing in my project).

But you're right it's unrelated, I'll remove it.

CHANGES.rst
```diff
@@ -5,7 +5,11 @@ master (unreleased)
 -------------------
 
 * Fix infinite recursion with multiple `MonitorField` and `defer()` or `only()`
-on Django 1.10+. Thanks Romain Garrigues. Merge of #242, fixes #241.
+  on Django 1.10+. Thanks Romain Garrigues. Merge of #242, fixes #241.
+* Fix issue #249 when extending `Manager` and `QuerySet` classes to add custom
```
carljm: Thanks for updating the changelog. Can we update this to actually describe the bug that was fixed in a way that someone who has experienced it would recognize without needing to look up the issue on GitHub? E.g. something like "* Fix InheritanceManager and SoftDeletableManager to respect `self._queryset_class` instead of hardcoding the queryset class. Merge of #250, fixes #249."

CHANGES.rst
```diff
@@ -5,7 +5,11 @@ master (unreleased)
...
+  methods.
+* Add mixins for `SoftDeletableQuerySet` and `SoftDeletableManager`, as stated
```
carljm: Thanks for noticing and fixing this!

CHANGES.rst
kezabelle: Just weighing in, though its only tangentially related.
Until 68bc61e82517cb01fa36c0d007091a8670425913 all references to GitHub issues in the changelog were "GH-<ticketNumber>"  historically. Carl, did you intentionally change to use the "#<ticketNumber>" form by preference for future changes, or should the unreleased ticket numbers go back to the old referencing format to be consistent? I don't think it matters much either way, short of someone knowing and grepping for "GH-[0-9]+", but as this PR touches both lines, rolling any change in might make sense here.

model_utils/tests/models.py
```diff
@@ -4,25 +4,24 @@
...
-
+from model_utils.tests.managers import CustomSoftDeleteManager
+from model_utils.tracker import FieldTracker, ModelTracker
 
 
 class InheritanceManagerTestRelated(models.Model):
     pass
 
 
-
```
kezabelle: I believe (@carljm correct me if I'm wrong) that Carl prefers/preferred 3 blank lines between classes intentionally, though that stance could've obviously changed.
Were it my choice, I'd omit the whitespace changes that follow subsequently in this file, as it only muddies the diff, but functionally it doesn't matter, and you're the one contributing herefor which we're thankful, so don't go changing it on my say so if Carl doesn't mind it :)

romgar: Indeed, that was @carljm preference. I respected the PEP8 when I refactored the tests though. Only because people are more used to PEP8 than anything else.

browniebroke: OK... Thanks for your comment. That's the result of "reformat code" in PyCharm (my editor) which by default, I think, follows PEP8.

I don't mind reverting these blank line changes, it's just a matter of being consistent over the whole project I guess (the `managers.py` IS following PEP8 and has 2 lines between classes).

Probably a separate discussion, but surely there must be an automated, customisable way of ensuring that, rather than relying on humans. Something that would be picked up by automated tests and ideally most editors.

browniebroke: Actually, I just realised that even in this same file, lines 185-300, there are 2 blank lines between classes.

CHANGES.rst
carljm: I don't have any strong feeling. I guess we could keep using `GH-` for consistency, but I also don't mind switching to the simpler `#`.

model_utils/tests/models.py
carljm: Years ago I used to prefer 3 blank lines between classes, but I've been following PEP8 for a long time now, just never did a full update of model-utils. PR welcome, or I'm fine letting it happen bit by bit. I don't really have much time for maintenance of model-utils these days, would be happy if someone wanted to take it over :-)

And yes, certainly we could add flake8 or pep8 or something to the Travis config, PR welcome for that too!

romgar: @carljm I will have more free time now (after this festive break) to assume some maintenance!

CHANGES.rst
```diff
@@ -5,7 +5,12 @@ master (unreleased)
 -------------------
 
 * Fix infinite recursion with multiple `MonitorField` and `defer()` or `only()`
-on Django 1.10+. Thanks Romain Garrigues. Merge of #242, fixes #241.
+  on Django 1.10+. Thanks Romain Garrigues. Merge of GH-242, fixes GH-241.
+* Fix `InheritanceManager` and `SoftDeletableManager` to respect `
```
carljm: seems like there's a stray backtick on the end of this line?

carljm:
Looks merge-ready to me, apart from the stray backtick

carljm:
Thanks! Merging.

Issue #251, opened 2017-01-09
Problem with STATUS in administration view. Show only first status.


error0608:
Having a model like this:
```
class ArticleAbstractModel(StatusModel, TimeStampedModel, TimeFramedModel):
    STATUS = Choices('published', 'draft')

    objects = LiveQuerySet.as_manager()

    class Meta:
        abstract = True
```

```
class ArticleModel(ArticleAbstractModel):
    title = models.CharField(max_length=100)
    slug = models.SlugField(unique=True)
    body = SplitField()
```
When I open this model (ArticleModel) on administrator page, it show me all objects with status == 'published'. Object with 'draft' status i don't see. But if do like this:
`STATUS = Choices('draft', 'published')`
System show me object only with 'draft' status.

Django == 1.10.5
django-model-utils == 2.6

romgar:
Hi @error0608 !

Can you show be the code behind `LiveQuerySet` class? Can you also copy paste your Admin model, if you customised it?

error0608:
Hi, @romgar
```
class LiveQuerySet(QuerySet):
    def live(self, statuses=['published']):
        today = timezone.now()
        return self.filter(status__in=statuses). \
            filter(Q(start__lte=today) | Q(start__isnull=True)). \
            filter(Q(end__gte=today) | Q(end__isnull=True))
```
```
@admin.register(ArticleModel)
class NewsAdmin(admin.ModelAdmin):
     pass
```

romgar:
@error0608 Thanks! I managed to reproduce it, I will have a deeper look.

romgar:
It seems that the Django Admin is using the `default_manager` to display the list of items in a model admin page.

But `StatusModel`, through the `add_status_query_managers` function, is adding custom managers (one per status), and these `add_to_class` calls seems to change the default manager.

I tried to save the default manager before the calls to `add_to_class`, and restore it after, it seems to solve this admin issue and keep the default manager.
Not sure that's the most elegant way to solve it though.

```
def add_status_query_managers(sender, **kwargs):
    """
    Add a Querymanager for each status item dynamically.

    """
    if not issubclass(sender, StatusModel):
        return

    # First, get current manager name...
    default_manager = sender._meta.default_manager

    for value, display in getattr(sender, 'STATUS', ()):
        if _field_exists(sender, value):
            raise ImproperlyConfigured(
                "StatusModel: Model '%s' has a field named '%s' which "
                "conflicts with a status of the same name."
[...]

PR #253, opened 2017-01-09
Add fix + tests for abstract manager inheritance issue with StatusModel

romgar:
I finally managed to reproduce the issue #251 in a test environment.

The conditions are quite precise: the issue appears when inheriting from an abstract model which redefines `objects` manager.

I guess it's related with the way Django orders the managers, which probably changed in `1.10`.

This solution prevents the default manager to be affected by `StatusModel` logic.


carljm:
Looks good to me, don't forgot to update CHANGES before merging.

PR #256, opened 2017-01-12
Drop old get_query_set syntax, replaced by get_queryset now

romgar:
Remove some artefacts left on #212:
- [ ] `get_query_set` usages

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -199,8 +199,6 @@ class InheritanceManagerMixin(object):
     def get_queryset(self):
         return self._queryset_class(self.model)
 
-    get_query_set = get_queryset
-
     def select_subclasses(self, *subclasses):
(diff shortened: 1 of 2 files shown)


romgar:
@carljm do you think we should also update the doc to only reflect current code? (like removing `PassThroughManager` mentions/examples, etc.). Versioning the doc would be ideal, but maybe not that easy...

carljm:
@romgar Yes, we should update the docs to remove outdated references. The docs in the repo should always match the code in the repo.

ReadTheDocs makes doc versioning very easy; I just went into the RTD admin and enabled building the docs for the latest bugfix release in each minor release series back to 1.5.0. I also added you as RTD admin for the project so you can manage this in future if you want.

romgar:
I didn't think it was that easy... I will manage it then, and clean the doc for the latest version at least.

Issue #61, opened 2013-07-22
Document alternative usages of PassThroughManager


treyhunner:
Example:

```
class CustomQuerySet(QuerySet):
     pass

class CustomManager(PassThroughManager):
    def __init__(self):
         return super(CustomManager, self).__init__(queryset_cls=CustomQuerySet)
```

This issue should be resolved at the same time as #56.

PR #259, opened 2017-01-18
Update documentation

romgar:
## Problem

The latest version of the documentation is not up-to-date, specially regarding `PassThroughManager`.

## Solution

Remove any reference of `PassThroughManager`, and update examples using them.

## Commandments

- [x] Cover it with tests.


romgar:
Will close #61, no more relevant.

romgar:
We could write other examples at some point.

PR #261, opened 2017-01-26
Add "license" arg to setup() call

kavdev:
## Problem

Services such as https://pyup.io grab license info from package metadata to display alongside each requirement. django-model-utils isn't providing that license info in it's metadata.

## Solution

Add a ``license`` argument to the ``setup()`` call in setup.py

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ def get_version(root_path):
 setup(
     name='django-model-utils',
     version=get_version(HERE),
+    license="BSD",
     description='Django model mixins and utilities',
     long_description=long_description,
     author='Carl Meyer',


kavdev:
This change doesn't add, alter, or remove functionality, so I figured a changelog entry and an issue isn't warranted.

kezabelle:
Arguably such services ought to parse the Trove classifiers for license information, too, because that's already in there, and according to [the docs on additional metadata](https://docs.python.org/3/distutils/setupscript.html#additional-meta-data) in an ideal world the license kwarg is for *indicating the license covering the package where the license is not a selection from the “License” Trove classifiers.*

In the real world, I don't see why this would be a problem to include.

kavdev:
fyi @jayfk

Issue #258, opened 2017-01-17
Remove tests from the package distribution


browniebroke:
## Problem 

Originally raised in #244, which was tracking the split of tests, but now closed. I'm in favour of getting the tests excluded from the distribution, I personally think testing code should not be present on Production machines. However, they seem to be needed to be able to package for RPM (see #230 @Germano0 ?)

The official packaging doc [suggests it](https://packaging.python.org/distributing/#packages), but it doesn't state it clearly.

Testing code is around 2/3 of the code in this repo (1031 lines vs. 2369 from a quick estimate), so that would also make the distribution lighter.

romgar:
I'm +1 to remove them from the distribution. I guess that, if we remove the test-related arguments in `setup.py` and exclude the `models_utils.tests` folder, it should be enough.

And ideally, it would have even been better to move the `tests` folder outside of the `model_utils` folder, and not even have to exclude anything in `setup.py`.

PR #264, opened 2017-02-15
Exclude tests from distribution

browniebroke:
## Problem

As explained in #258, this is to exclude the tests from the distributed version of this library.

## Solution

- Move the `tests` folder outside of the main package.
- Exclude them from `find_package()` in setup.py
- Update test command to call `runtests.py` directly (rather than `setup.py test`)
- Updated paths accordingly in test runner and translations generator script

## Commandments

- [x] Cover it with tests.


romgar:
Thanks for the MR @browniebroke ! Did you check that the number of tests running has not changed with your update?

browniebroke:
Good point. Forgot about that.

I compared just now, against the latest build on master (549). I looked at 2 environments: Python 2.7 with Django 1.8 and Python 3.5 with Django 1.10 and they are all running 256 test cases.

tox.ini
```diff
@@ -20,4 +20,4 @@ deps =
     django_trunk: https://github.com/django/django/tarball/master
     freezegun == 0.3.8
 
-commands = coverage run -a setup.py test
```
romgar: Any reason to remove the `-a` option?

tests/models.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import unicode_literals
+from __future__ import unicode_literals, absolute_import
```
romgar: Do we really need this `absolute_import`?

browniebroke: Yes, it's failing to import `MutableField` on Python 2.7 without it.

tox.ini
browniebroke: That's a mistake from myself, will add it back.

romgar:
Thanks @browniebroke !!

PR #265, opened 2017-02-15
Run tests for Django 1.11

browniebroke:
## Problem

The first alpha for Django 1.11 is out, and the feature are now frozen as per [the roadmap](https://code.djangoproject.com/wiki/Version1.11Roadmap). This adds it to the build matrix.
Since Django's master branch has already dropped support for Python 2, we don't need to test Python 2.7 against it anymore.

## Solution

- Added new environments to Toxfile + Travis, including 1.11 + Python 3.6.
- Marked as allowed failures, tests don't pass yet.
- Removed Python 2.7 vs. Django master

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
   - TOXENV=py27-django19
   - TOXENV=py27-django110
   - TOXENV=py27-django111
-  - TOXENV=py27-django_trunk
   - TOXENV=py33-django18
   - TOXENV=py34-django18
   - TOXENV=py34-django19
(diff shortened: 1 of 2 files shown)


browniebroke:
I'm not entirely happy with the setup for Python 3.6, as it's either:
- Not finding the interpreter
- Running twice (when adding include python 3.6 in the matrix)
- Requires an include + exclude in the matrix

[Django REST framework](https://github.com/tomchristie/django-rest-framework) is using tox-travis to do it, but it's probably more verbose.

browniebroke:
Not needed anymore, Matrix has been updated in another PR.

PR #286, opened 2017-07-16
Fix excess number of queries

### LINKED ISSUE #274: Excess number of queries generated with `get_subclass`
## Problem

After upgrading from 2.6.1 to 3.0.0 the number of queries generated by a `get_subclass` increases by (on average) half the number of subclasses there are.

## Environment

- Django Model Utils version: 3.0.0
- Django version: 1.10.7
- Python version: 3.4

## Code examples

Example `models.py` file
```
from django.db import models
from model_utils.managers import InheritanceManager


class BaseDemo(models.Model):
    my_field = models.IntegerField()
    objects = InheritanceManager()


class DemoA(BaseDemo):
    a_field = models.IntegerField()


class DemoB(BaseDemo):
    b_field = models.IntegerField()


class DemoC(BaseDemo):
    c_field = models.IntegerField()
```

Then in a Django shell I create an instance of each type and use `BaseDemo` to get a subclass:
```
from demo.models import *
DemoA.objects.get_or_create(my_field=1, a_field=2)
DemoB.objects.get_or_create(my_field=1, b_field=3)
DemoC.objects.get_or_create(my_field=1, c_field=4)
BaseDemo.objects.get_subclass(id=3)
```

In `django-model-utils==2.6.1` this would generate the following query and proper result:
```
SELECT "demo_basedemo"."id", "demo_basedemo"."my_field", "demo_demoa"."basedemo_ptr_id", "demo_demoa"."a_field", "demo_demob"."basedemo_ptr_id", "demo_demob"."b_field", "demo_democ"."basedemo_ptr_id", "demo_democ"."c_field" FROM "demo_basedemo" LEFT OUTER JOIN "demo_demoa" ON ("demo_basedemo"."id" = "demo_demoa"."basedemo_ptr_id") LEFT OUTER JOIN "demo_demob" ON ("demo_basedemo"."id" = "demo_demob"."basedemo_ptr_id") LEFT OUTER JOIN "demo_democ" ON ("demo_basedemo"."id" = "demo_democ"."basedemo_ptr_id") WHERE "demo_basedemo"."id" = 3; args=(3,)
```

with `django-model-utils==3.0.0` this generates the correct query, then an additional query for each sub-model-class:
```
SELECT "demo_basedemo"."id", "demo_basedemo"."my_field", "demo_demoa"."basedemo_ptr_id", "demo_demoa"."a_field", "demo_demob"."basedemo_ptr_id", "demo_demob"."b_field", "demo_democ"."basedemo_ptr_id", "demo_democ"."c_field" FROM "demo_basedemo" LEFT OUTER JOIN "demo_demoa" ON ("demo_basedemo"."id" = "demo_demoa"."basedemo_ptr_id") LEFT OUTER JOIN "demo_demob" ON ("demo_basedemo"."id" = "demo_demob"."basedemo_ptr_id") LEFT OUTER JOIN "demo_democ" ON ("demo_basedemo"."id" = "demo_democ"."basedemo_ptr_id") WHERE "demo_basedemo"."id" = 3; args=(3,)
SELECT "demo_basedemo"."id", "demo_basedemo"."my_field", "demo_demoa"."basedemo_ptr_id", "demo_demoa"."a_field" FROM "demo_demoa" INNER JOIN "demo_basedemo" ON ("demo_demoa"."basedemo_ptr_id" = "demo_basedemo"."id") WHERE "demo_demoa"."basedemo_ptr_id" = 3; args=(3,)
SELECT "demo_basedemo"."id", "demo_basedemo"."my_field", "demo_demob"."basedemo_ptr_id", "demo_demob"."b_field" FROM "demo_demob" INNER JOIN "demo_basedemo" ON ("demo_demob"."basedemo_ptr_id" = "demo_basedemo"."id") WHERE "demo_demob"."basedemo_ptr_id" = 3; args=(3,)
SELECT "demo_basedemo"."id", "demo_basedemo"."my_field", "demo_democ"."basedemo_ptr_id", "demo_democ"."c_field" FROM "demo_democ" INNER JOIN "demo_basedemo" ON ("demo_democ"."basedemo_ptr_id" = "demo_basedemo"."id") WHERE "demo_democ"."basedemo_ptr_id" = 3; args=(3,)
```

I say the number of calls increases by (on average) half the number of subclasses because they are sorted and it only generates queries up-to the point it finds the proper subclass. So if I did `BaseDemo.objects.get_subclass(id=1)`, it would generate the correct query and a single follow up query.

sk1p:
## Problem

This fixes #274, which is a regression on Django 1.9/1.10 causing many unnecessary database queries, because the ``_get_sub_obj_recurse`` logic is run twice (once on the parent instance and again on the subclass instance). The cached values from the ``select_related(...)`` call in ``select_subclasses()`` get lost when selecting the subclass instances, and the database is hit again, _for ~ each item in the queryset!_

## Solution

On Django >= 1.9, ``InheritanceQuerySetMixin.iterator()`` now yields from ``InheritanceIterable`` and doesn't re-run the ``_get_sub_obj_recurse`` logic.

## Commandments

- [x] Cover it with tests.


sk1p:
I also have a [less minimal patch](https://github.com/jazzband/django-model-utils/compare/master...sk1p:excess_queries_2) that removes the duplicated code, if that is preferred.

plinton:
@tony any chance you could take a look at this one? I can confirm that we are using the "less minimal patch" in production and it appears to be working well

carljm:
The added test looks good here, and if tests pass, seems fine to me. I'm probably not the most qualified reviewer, however, having not really used the Django ORM in production since 1.8.

tests/test_managers/test_inheritance_manager.py
```diff
@@ -115,6 +115,14 @@ def test_prior_select_related(self):
                 "inheritancemanagertestchild2").get(pk=self.child1.pk)
             obj.inheritancemanagertestchild1
 
+    def test_select_subclass_num_queries(self):
+        with self.assertNumQueries(1):
+            objs = self.get_manager().select_subclasses(
+                "inheritancemanagertestchild1",
+                "inheritancemanagertestchild2",
+            )
+            list(objs)
```
tony: If the test is conditional on django version, is there anything we want to assert on a django version basis here?

tests/test_managers/test_inheritance_manager.py
```diff
@@ -116,6 +116,17 @@ def test_prior_select_related(self):
...
+
+    def test_version_determining_any_depth(self):
+        self.assertIsNone(self.get_manager().all()._get_maximum_depth())
```
Natim: It seems that `_get_maximum_depth()` was removed with https://github.com/jazzband/django-model-utils/commit/0fc0b44c95de3e39b71a1e01a3ef36a6bc887d12#diff-a4a84bac70b6f91df5344a194dbe270aL233

tests/test_managers/test_inheritance_manager.py
Natim: @tony I guess the fact that tox is testing with all django and python supported version helps to make sure the test passes with all supported configuration.

Natim:
Superseeded by #389

PR #310, opened 2018-01-19
Removed missing include files in MANIFEST.in

Germano0:
## Problem
I was updating django-model-utils on Fedora to the  3.1.1 version, when I got some warnings concerning missing include files in MANIFEST.in
https://kojipkgs.fedoraproject.org//work/tasks/1700/24281700/build.log

## Solution
I removed missing entries from the file


RDIL:
I’ll fix it.

jezdez:
@RDIL Hey, the patch isn't correct, now it only includes the local files for the German language, not the others anymore. Could you please fix this?

PR #317, opened 2018-04-03
Fix handling of deferred fields on django 1.10+

### LINKED ISSUE #278: AttributeError: 'CountryDescriptor' object has no attribute 'field_name'
## Problem

When a `CountryField` is used as a deferred field, `FieldTracker` breaks at the following line:

                    field_tracker = DeferredAttributeTracker(
                        field_obj.field_name, None)

## Environment

- Django Model Utils version: 3.0.0
- Django version: 1.11.2
- Python version: 3.5.2
- Django Countries version: 4.6.1

## Code examples

models.py
```
from django.db import models
from django_countries.fields import CountryField
from model_utils import FieldTracker


class Location(models.Model):
    country = CountryField(blank=True)

    tracker = FieldTracker()
```

The following code fails on Django 1.11 but works on Django 1.9:
```
Location.objects.create()
print(list(Location.objects.only('id')))
```

Here is a test project with a single test that fails:
[django_model_utils_countries_tracker_test.zip](https://github.com/jazzband/django-model-utils/files/1097279/django_model_utils_countries_tracker_test.zip)

lucaswiman:
Fixes #278. cc @utapyngo @djstein 

## Problem

Prior to Django 1.10, deferred attributes were handled by constructing a new model class with custom descriptors for the deferred attributes. After 1.10, deferred fields are tracked by whether the attribute is present in the instance `__dict__`.

`FieldTracker` handled tracking on these fields by overwriting the descriptors on the model class. This means that in 1.10+, the instance model class _is_ the model, so `FieldTracker` can introduce changes to the base class whenever deferred fields are used that persist on other queries. #278 This is reproduced in a test case.

## Solution

I updated the `finalize_class` method to _wrap_ field descriptors. This preserves the custom descriptor behavior, and means that deferring fields does not lead to permanent changes in base model classes. This is only done for django versions after 1.10: the previous behavior is preserved for 1.8 and 1.9.

Since the relevant code branches behavior on unsupported versions of Django, I also removed all branching behavior for versions of django listed as unsupported in the CHANGELOG.

## Commandments

- [x] Cover it with tests. _Reproduced in test cases, which fail prior to 54cc1507a79460497d40f6cba779f67a4b5f8041 and pass afterwards._
- [x] Pay attention to backward compatibility, or if it breaks it, explain why. _The only backwards incompatible change is that the `tracker._deferred_fields` attribute is not set on django versions >=1.10. Since `_deferred_fields` is a private attribute, I wouldn't consider this a breaking change._
- [x] Update documentation (if relevant). _I don't think it is relevant.

Regarding PEP8, I ran `flake8`, and it generated a number of errors. I think my code didn't introduce new PEP8 violations, but I'm not sure. If you want, I can submit a separate PR fixing all the PEP8 violations, and rebase onto this branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 - 3.6
 install: pip install tox-travis codecov
 # positional args ({posargs}) to pass into tox.ini
-script: tox -- --cov
+script: tox -- --cov --cov-append
 after_success: codecov
 deploy:
   provider: pypi
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -43,3 +43,4 @@ Trey Hunner <trey@treyhunner.com>
 Karl Wan Nan Wo <karl.wnw@gmail.com>
 zyegfryed
 Radosław Jan Ganczarek <radoslaw@ganczarek.in>
+Lucas Wiman <lucas.wiman@gmail.com>
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ CHANGES
 
 master (unreleased)
 -------------------
+- Fix handling of deferred attributes on Django 1.10+, fixes GH-278
 
 3.1.2 (2018.05.09)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,8 +3,12 @@ CHANGES
 
 master (unreleased)
 -------------------
+
+3.1.2 (2018.05.09)
+------------------
 * Update InheritanceIterable to inherit from
   ModelIterable instead of BaseIterable, fixes GH-277.
+
 
 3.1.1 (2017.12.17)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,15 +4,13 @@ CHANGES
 master (unreleased)
 -------------------
 - Fix handling of deferred attributes on Django 1.10+, fixes GH-278
+- Fix `FieldTracker.has_changed()` and `FieldTracker.previous()` to return
+  correct responses for deferred fields.
 
 3.1.2 (2018.05.09)
 ------------------
 * Update InheritanceIterable to inherit from
   ModelIterable instead of BaseIterable, fixes GH-277.
-
-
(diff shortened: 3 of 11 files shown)


lucaswiman:
😬 I'll investigate why code coverage went down by so much.

lucaswiman:
I fixed a few coverage issues, but I think maybe the codecov integration isn't running the full test suite? All of the code which only runs on older versions of django shows as "not covered". I think this is an artifact of the coverage setup, and not reflective of an inadequacy in the tests I added.

lucaswiman:
Hmm, I wish I'd noticed #313 existed before starting on this. cc @jcushman

I think we're fixing different bugs with the handling of deferred fields, but the patches are slightly incompatible.

lucaswiman:
@jcushman I think this now includes fixes from #313, updated to work on Django 1.10+. Could you take a look?

@carljm What is required to get this merged and deployed this to pypi?

lucaswiman:
OK, I got the codecov checks passing by correctly aggregating the different test environments and adding a few assertions to the tests.

jcushman:
Hi! I don't have my head super in this, but I see that my tests are there and passing, which is great. 👍

It sounds like you can join Jazzband and merge this yourself, if you're comfortable with the code at this point: https://github.com/jazzband/django-model-utils/pull/279#issuecomment-386069220

Alternatively you might @ a few folks who have merged PRs in the last year for another pair of eyes.

jcushman:
(Oh one nitpick, looks like the merge on CHANGES.rst put a line in the wrong place.)

lucaswiman:
> you might @ a few folks who have merged PRs in the last year for another pair of eyes.

@tony Would you care to take a look?

tony:
@jcushman @lucaswiman I don't use the deferred fields part of django-model-utils. 

I breezed through the code and it LGTM. I think that's the best I can offer.

I don't want to offer more since I don't have a Django website using that part of django-model-utils to test on fully.

lucaswiman:
Thanks @tony and @jcushman! I'll merge this and fix pep8 in another PR.

PR #319, opened 2018-04-19
✅ Ensure TimeStampedModel modified equals created on initial creation.

### LINKED ISSUE #247: TimeStampedModel improvement ideas
Hi I think the behavior of the TimeStampedModel class is a little misleading / less useful because of the following:

- modified is set even though the model was only saved once (when creating).
- modified is different than created even though they were set at the same time.

I propose two possible changes to the TimeStampedModel so it can be possible to tell whether or not a model was only created or created and later modified/updated.

1. modified is set to null on create and only changed on subsequent saves.
2. modified is set the the exact same value as created on create.

It seems like 1. is a better solution to me because that way the modified field and the AutoModifiedField aren't tied directly to the AutoCreatedField also existing on the model.

andrlik:
## Problem

Currently any model that descends from TimeStampedModel is difficult to query for edits. Upon instance creation, the modified timestamp is set to be a fraction of a second later than the created. This makes it impossible to tell if a model was edited, or simply has the timestamp from creation.

## Solution

Add logic to AutoLastModifiedField which checks to see if the associated created field of the correct cField class type is also present. If the instance has not yet been saved (missing a pk), then set the value to modified to be equal to created.

Fixes #247


andrlik:
Good points. I've updated the PR with those changes.

PR #320, opened 2018-04-22
Fixes a bug with subclasses missing from a cloned qs

femdom:
## Problem

Cloned `InheritanceQuerySet` misses `subclasses` from the original one.

```
In [37]: qs = Contact.objects.select_subclasses()

In [38]: qs
Out[38]: <InheritanceQuerySet [<EmailContact: EMail contact: ***redacted***>]>

In [39]: qs._clone()
Out[39]: <InheritanceQuerySet [<Contact: Contact object (1)>]>
```

## Solution

`_clone` implementation should copy `subclasses` attribute from the original queryset.

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -99,7 +99,10 @@ class InheritanceQuerySetMixin(object):
 
     def _clone(self, klass=None, setup=False, **kwargs):
         if django.VERSION >= (2, 0):
-            return super(InheritanceQuerySetMixin, self)._clone()
+            qs = super(InheritanceQuerySetMixin, self)._clone()
+            if hasattr(self, 'subclasses'):
+                qs.subclasses = self.subclasses
+            return qs
 
         for name in ['subclasses', '_annotated']:
             if hasattr(self, name):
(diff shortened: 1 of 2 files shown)


model_utils/managers.py
```diff
@@ -99,7 +99,10 @@ def _chain(self, **kwargs):
 
     def _clone(self, klass=None, setup=False, **kwargs):
         if django.VERSION >= (2, 0):
-            return super(InheritanceQuerySetMixin, self)._clone()
+            qs = super(InheritanceQuerySetMixin, self)._clone()
+            if hasattr(self, 'subclasses'):
+                qs.subclasses = self.subclasses
```
lucaswiman: Is there a reason to not move the copying logic from lines 107-109 above the version check? In particular, this change doesn't account for `_annotated`.

illing2005: Problem is that in django2 and above queryset does not allow `kwargs` in the clone function. 
Therefore the copying logic is a bit different. 
I created a new pull-request #335 to resolve this also for the `_annotated` field.

PR #324, opened 2018-04-30
Throw an error in FieldTracker when using untracked field with .previous

furious-luke:
## Problem

Previously if `previous` was called on a field tracker and the field was not being tracked it would return `None`. Having consistent behavior between `previous` and `has_changed` would seem to make sense, but there may be a reason why this is not so?

## Solution

I've moved the `FieldError` exception into `previous` from `has_changed`; `has_changed` calles `previous` so the exception will still be used from `has_changed`.

## Commandments

- [ ] Cover it with tests.

I can add a test for this if you think it's worth merging?

### DIFF
--- a/model_utils/tracker.py
+++ b/model_utils/tracker.py
@@ -66,14 +66,14 @@ class FieldInstanceTracker(object):
 
     def has_changed(self, field):
         """Returns ``True`` if field has changed from currently saved value"""
-        if field in self.fields:
-            return self.previous(field) != self.get_field_value(field)
-        else:
-            raise FieldError('field "%s" not tracked' % field)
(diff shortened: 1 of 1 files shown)


lucaswiman:
@furious-luke The test failures look related to these changes.

model_utils/tracker.py
```diff
@@ -66,14 +66,14 @@ def current(self, fields=None):
 
     def has_changed(self, field):
         """Returns ``True`` if field has changed from currently saved value"""
-        if field in self.fields:
```
Natim: @furious-luke shall we keep the if in both methods? Can you add a test for your change?

PR #329, opened 2018-07-02
Fix pep8 violations

lucaswiman:
@tony @carljm 
## Problem

One of the checkboxes in the PR checklist ("Write PEP8 compliant code.") wasn't covered in #317, since there were many pep8 violations already present in the repo. I said I would fix flake8 violations in another pull request.

## Solution

I added a `flake8` tox environment which runs flake8 to enforce pep8-style code formatting. I've ignored the following errors:

### E731 do not assign a lambda expression, use a def
Personally this one always struck me as sort of silly and pedantic. I don't think it improves code readability, and can sometimes detract from it. I can fix the violation of this rule if requested.

### W503 line break before binary operator
I think this actually leads to less readable code in many cases, since violating this allows showing the operator connecting the two lines at the beginning of the line (emphasized) rather than the end (deemphasized). As before, I can fix the violation of this rule if requested.

### E402 module level import not at top of file

This seems to have been a false positive. I think `flake8` dislikes this conditional import, but it looks fine to me:
https://github.com/jazzband/django-model-utils/blob/16dec4d12d082aeeb15e7d30c85bd1bf8e33e584/model_utils/models.py#L7-L15

If requested, I can either exclude this particular set of imports, or alter the ordering of execution, though it doesn't seem like a win for code clarity.

### E501 line too long
There are a large number of violations of the 80 character line length limit, which argues against fixing this. In particular, it would likely introduce difficult merge conflicts with most of the open pull requests.

One option would be to pick a longer limit, like the 88 characters preferred by [Black](https://github.com/ambv/black#line-length), or some other higher value like 100 or 120. The longest line is 114 characters, so choosing 120 would require no code changes. Choosing 100 would require few code changes (6 lines).

### DIFF
--- a/model_utils/__init__.py
+++ b/model_utils/__init__.py
@@ -1,4 +1,4 @@
-from .choices import Choices
-from .tracker import FieldTracker, ModelTracker
+from .choices import Choices  # noqa:F401
+from .tracker import FieldTracker, ModelTracker  # noqa:F401
 
 __version__ = '3.1.2'
--- a/model_utils/choices.py
+++ b/model_utils/choices.py
@@ -92,7 +92,7 @@ class Choices(object):
                     raise ValueError(
                         "Choices can't take a list of length %s, only 2 or 3"
                         % len(choice)
-                        )
+                    )
             else:
                 store((choice, choice, choice))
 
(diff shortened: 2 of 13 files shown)


lucaswiman:
Seems like there are no objections, and this seems like an incremental improvement, so I'm going to merge this. The line length thing can be figured out later.

Issue #348, opened 2018-11-13
Is there interest in creating a more effective way of doing IN


jonathan-s:
## Feature
So in the following answer on [stackexchange](https://dba.stackexchange.com/a/91254/132631) it details how using `IN` can be quite ineffective when you are doing a query with a huge array using `IN`. Currently there is no elegant way of doing this directly in django. 

I've been writing a custom queryset that solves this though. Would there be interest in having this queryset to be part of this package?

PR #351, opened 2018-11-24
Add a JoinManager that helps with performance

jonathan-s:
## Problem

Implementing the suggestion as described in #348. 

## Description

The PR does the following things
- Updates coveragerc to include all relevant files, not all files were included before. Hence codecoverage complaining
- Replace sqlite3 with postgres for better reliability (also you can't run several statements in a cursor in sqlite3 which is another reason for opting for postgres). 
- Implementing the suggestion in #348 
- Dropping django 1.8. The field `rel`, was not available and django 1.8 is way out of date anyway so it's better to drop it all together. 

Solution in this PR. Instead of using sqlite, it's also using postgres. 

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,2 @@
 [run]
-source = model_utils
-omit = .*
-       tests/*
-       */_*
+include = model_utils/*.py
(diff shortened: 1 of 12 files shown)


jonathan-s:
@jordan-dimov and @auvipy Perhaps you'd be interested in reviewing this?

jonathan-s:
🎉

PR #360, opened 2019-02-26
Add UUIDModel and UUIDField

### LINKED ISSUE #332: UUID Model
Would there be any interest in adding a mixin to this library for models with UUID PKs? This is a common requirement for any application that exposes PKs in any public setting.

```python
class UUIDModel(models.Model):
    """
    An abstract base class model with a UUID as its primary key.
    """
    id = models.UUIDField(
        primary_key=True,
        default=uuid.uuid4,
        editable=False,
    )

    class Meta:
        abstract = True
```

marfyl:
## Problem

This fix #332 and the interest to have a UUIDModel that override pk primary key with an uuid field.

## Solution

This abstract base class UUIDModel provides id field on any model that inherits from it which will be the primary key.

Also implements an UUIDField so If you dont want to set id as primary key or change the field name, you can do it overriding it with the custom field.

You can set the uuid version. Versions 1,3,4 and 5 are supported, default is 4.

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -50,4 +50,4 @@
 | Jack Cushman <jcushman@law.harvard.edu>
 | Zach Cheung <kuroro.zhang@gmail.com>
 | Daniel Andrlik <daniel@andrlik.org>
-
+| marfyl <github.com/marfyl>
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ master (unreleased)
 - Fix patched `save` in FieldTracker
 - Upgrades test requirements (pytest, pytest-django, pytest-cov) and 
   skips tox test with Python 3.5 and Django (trunk)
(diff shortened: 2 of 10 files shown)


marfyl:
Hi guys, what do you think about merge this? I use it a lot in my projects as a mixin inheriting from the class TimeStampModel. I hope it brings value to the community

auvipy:
it's a pretty big change, so wait for a more rigorous review.

marfyl:
Ok, if you need to change something of the approach during the code review, just write me in PR

auvipy:
sure

marfyl:
did you have a chance to review the PR?

auvipy:
not yet. could you increase the test coverage a bit?

marfyl:
done

marfyl:
when do u expect to launch a new release??

PR #363, opened 2019-03-11
managers: honor OneToOneField.parent_link=False

Hyask:
## Problem

If a child model needs a `OneToOneField` with any of its parent class that use `InheritanceManager` along with the usual `parent_link` generated by Django, `model_utils` adds by mistake the reverse link of this relation to the usual list of possible subclasses. Depending of the length of the classes names, and the depth of the class tree, a bug may be triggered or not, resulting in a child object being cast to the wrong class (basically one with a longer name).

## Solution

When building the list of possible subclasses, check that `parent_link` is `True` before adding the class to the list.

## Commandments

- [x] Cover it with tests.
- [ ] ~~Update documentation (if relevant).~~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ CHANGES
 
 master (unreleased)
 -------------------
+- Honor `OneToOneField.parent_link=False`.
 - Fix handling of deferred attributes on Django 1.10+, fixes GH-278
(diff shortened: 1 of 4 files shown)


Hyask:
@auvipy Thanks for the merge!
Do you have an idea when this fix will released (at least on PyPi)? I've seen there hasn't been any release for quite some time now, can we expect one coming in the near future?

Issue #331, opened 2018-08-03
Allow FieldTracker on Django user model (and other children of abstract base classes)


jcushman:
## Problem

Here is a branch with failing tests that I think should work (and that I think worked in earlier versions?): https://github.com/jazzband/django-model-utils/compare/master...jcushman:abstract-test-failure

This seems to be related to the stuff that @lucaswiman added in #317. @lucaswiman, I'm hoping based on that work you might have some clever idea for how to fix this. :)

Here's the situation: you have an abstract base class that defines a static attribute like `is_active = True`, and a concrete model inheriting from that class that defines a field like `is_active = models.BooleanField(default=True)`. The model then throws an error on `save()`:

```
from django.contrib.auth.models import AbstractUser

class MyUser(AbstractUser):
    tracker = FieldTracker()

MyUser().save()
```

Result:

```
    def __get__(self, instance, owner):
        if instance is None:
            return self
        was_deferred = self.field_name in instance.get_deferred_fields()
>       value = self.descriptor.__get__(instance, owner)
E       AttributeError: 'bool' object has no attribute '__get__'

model_utils/tracker.py:43: AttributeError
```

It would be great for this to work, because tracking changes on the Django user model is handy.

Debugging that error, I found the problem boils down to this:

```
class AbstractUser(models.Model):
    is_active = True

    class Meta:
        abstract = True

class MyUser(AbstractUser):
    is_active = models.BooleanField(default=True)
    tracker = FieldTracker()

MyUser().save()
```

The reason that fails is https://github.com/jazzband/django-model-utils/blob/master/model_utils/tracker.py#L218 :

```
descriptor = getattr(sender, field_name)
...
setattr(sender, field_name, wrapped_descriptor)
```

... which boils down to setting `MyUser.is_active = DescriptorWrapper(MyUser.is_active)`. And _that_ doesn't work because you expect `MyUser.is_active` to start with a value of `DeferredAttribute('is_active')`, but it actually returns `True`. For reasons I don't understand, when you override a static attribute on an abstract class, you get the base class's value back instead of the subclass's.

I tried tweaking tracker.py with variations on `descriptor = getattr(sender, field_name) if field_name in sender.__dict__ else DeferredAttribute(field_name)`, but that broke foreign key fields and feels pretty janky anyway.

Any ideas on how to get this working?

Thanks!

## Environment

- Django Model Utils version: master
- Django version: 2.1
- Python version: 3.6
- Other libraries used, if any:

[...]

PR #370, opened 2019-05-01
Fix/deferred attribute assignment

jarekwg:
Reverts #367
Adds a test for #331

PR also sneaks in some linter configs.

Issue not fixed here. I suspect it's a django issue. Findings reported in #331.

### DIFF
--- a/model_utils/tracker.py
+++ b/model_utils/tracker.py
@@ -30,7 +30,6 @@ class DescriptorMixin(object):
 
 
 class DescriptorWrapper(object):
-
     def __init__(self, field_name, descriptor, tracker_attname):
         self.field_name = field_name
         self.descriptor = descriptor
@@ -40,10 +39,7 @@ class DescriptorWrapper(object):
         if instance is None:
             return self
         was_deferred = self.field_name in instance.get_deferred_fields()
-        try:
-            value = self.descriptor.__get__(instance, owner)
-        except AttributeError:
-            value = self.descriptor
+        value = self.descriptor.__get__(instance, owner)
         if was_deferred:
             tracker_instance = getattr(instance, self.tracker_attname)
             tracker_instance.saved_data[self.field_name] = deepcopy(value)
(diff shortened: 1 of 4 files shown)


jarekwg:
Yeah these failures are expected. Going to try see if the fix can be made in django itself. Just made PR now, to include failing test.
I guess, for the time being, the most useful fix would be #336, but it comes with its own edge cases that it fails to cover...

Natim:
can you add a isort rule to the makefile too?

model_utils/tracker.py
```diff
@@ -40,10 +39,7 @@ def __get__(self, instance, owner):
         if instance is None:
             return self
         was_deferred = self.field_name in instance.get_deferred_fields()
-        try:
-            value = self.descriptor.__get__(instance, owner)
-        except AttributeError:
-            value = self.descriptor
+        value = self.descriptor.__get__(instance, owner)
```
Natim:
```suggestion
        try:
            value = self.descriptor.__get__(instance, owner)
        except AttributeError:
            value = self.descriptor
```

jarekwg: Hey, this just reverts my changes?
I've already forgotten a lot of what was going on here, but I believe the try-catch was only a bandaid fix that caused other edge cases to fail.
The real fix, I believe, needs to be made in django. See my open ticket and PR with them here: https://code.djangoproject.com/ticket/30427#ticket
Helping get that merged in would, then allow us to fix the issue properly here.

Natim: Ok thank you for your advice.

jarekwg:
> can you add a isort rule to the makefile too?

I've not used makefiles this way before, but can take a look when it's time to finally look into getting this merged in. I think we're blocked on django currently.

jarekwg:
Closing. Idk if this was addressed in the end by someone else somewhere else, but I don't have the resources to delve into here again.. ;/

PR #373, opened 2019-06-21
Release 3.2.0.

### LINKED ISSUE #366: New release?
Is there a new release planned in near future? 3.1.2 is from May 2018 and master changed a lot since that

charettes:
Planning on tagging this change with the `3.2.0` to kickoff the release process once it's merged.

Looks like version number was already bumped in https://github.com/jazzband/django-model-utils/commit/05671695bb347fea57fcf33c2f0897d918184819 but never release, not sure why.

Fixes #366

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 CHANGES
 =======
 
-master (unreleased)
+3.2.0 (2019.06.21)
 -------------------
 - Catch `AttributeError` for deferred abstract fields, fixes GH-331.
 - Update documentation to explain usage of `timeframed` model manager, fixes GH-118


cristeaadrian:
While this is great, it's still not available on PyPI.

charettes:
@cristeaadrian upload to Pypi is unfortunately not something I've done.

Based on [Jazzband release guidelines](https://jazzband.co/about/releases) the package now requires approval from @carljm [as project lead](https://jazzband.co/projects/django-model-utils) since it has been uploaded to Jazzband's servers.

I've applied to become a second lead for this project (https://github.com/jazzband/roadies/issues/151) and tried pinging people on IRC to get things going but until a lead approves the release it won't be on Pypi.

cristeaadrian:
@charettes Ah, I see. Thanks for the quick response, I hope your application is succesful since I'm in the same boat as you (employer depends on this package).

carljm:
Approved release, looks like 3.2.0 is now on PyPI. Commented in favor of adding @charettes as project lead, AFAICS in the jazzband UI I don't actually have the ability to make that happen though.

charettes:
@carljm thanks for chiming in and the clarifications regarding publishing. I was under the impression that since the package got uploaded to jazzband.co there would be some UI to approve and publish the package.

carljm:
@charettes Oh, I was able to release the 3.2.0 package, no problem there!

What I'm not able to do myself (AFAICT) is add you as a project lead so you can do it in future. But I commented in favor of adding you, hopefully a jazzband roadie is able to make that happen. Thanks for stepping up to help with maintenance of this package!

charettes:
Alright that makes sense, thanks for everything @carljm!

PR #381, opened 2019-08-05
Fix tests

Atorich:
## Problem

I have found that `AbstractModelTrackerTests` is not quite correct.
Since it inherited from `FieldTrackerTestCase` that just implements some extra assertions, there are no actual tests for this case.

## Solution

I have fixed `AbstractModelTrackerTests` and it seems that some tests are failed for the case with models inherited from some AbstractModel with fields defined (like AbstractUser and its children)

### DIFF
--- a/tests/test_fields/test_field_tracker.py
+++ b/tests/test_fields/test_field_tracker.py
@@ -1,5 +1,5 @@
 from __future__ import unicode_literals
-
+from unittest import skip
 import django
 from django.core.exceptions import FieldError
 from django.test import TestCase
@@ -787,6 +787,7 @@ class InheritedModelTrackerTests(ModelTrackerTests):
         self.assertTrue(self.tracker.has_changed('name2'))
 
 
+@skip
 class AbstractModelTrackerTests(ModelTrackerTests):
 
     tracked_class = TrackedAbstract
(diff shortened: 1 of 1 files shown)


Atorich:
Just closed without any comments?

Natim:
The PR says Fix tests, but tests work on master and the CI status is a failing tests.
The PR is one line change and break the tests. I don't know what more to tell.

Can you elaborate on what you are trying to do?

Atorich:
Have you read the description?
AbstractModelTrackerTests does not test something -- that's why "tests work".
One line change -- is something that actually runs the tests which are failing, you're right.

Natim:
Feel free to fix the tests and to reopen if you feel like something needs to be done.

Atorich:
There is nothing to fix in the tests. 
The problem with using FieldTracker on models marked as `abstract=True` is still here and my PR just proves it.

I think it is better to keep the test case from my PR but `@skip` it for now because of further investigation [here](https://github.com/jazzband/django-model-utils/pull/367
) and [here](https://github.com/jazzband/django-model-utils/pull/370) means that the problem is out of the scope of `model-utils`.
But leaving `ModelTrackerTests` is confusing.

Natim:
Agreed

PR #384, opened 2019-08-19
Added `Choices().subset()`.

Asday:
## Problem

Throughout a project I'm working on, there are lines such as this:

```python3
failure_reason = forms.ChoiceField(
    ...,
    choices=tuple(Model.CHOICES)[1:3],
)
```

Which are unreadable and fragile.

## Solution

With this PR, they can be changed to:

```python3
failure_reason = forms.ChoiceField(
    ...,
    choices=Model.CHOICES.subset('cancelled', 'failed', 'aborted')
)
```

In the mean time, we could subclass `Choices`, or write a helper function, but I figured I'd throw this together, as it seems like an entirely reasonable use of `Choices()`, that very well may help others.

That's the only change - the new method.  (And surrounding tests, documentation, etc.)

(I got to the part of `CONTRIBUTING.rst` that mentions I have to update `CHANGES.rst`, and got a little confused, so I also bumped the version.  Not certain if I did it right, or did the right thing).

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -1,4 +1,5 @@
 | ad-m <github.com/ad-m>
+| Adam Barnes <sara.and.zuka+django-model-utils@gmail.com>
 | Alejandro Varas <alej0varas@gmail.com>
 | Alex Orange <crazycasta@gmail.com>
 | Alexey Evseev <myhappydo@gmail.com>
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 CHANGES
 =======
 
+3.3.0 (2019.08.19)
+------------------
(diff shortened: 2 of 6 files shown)


Asday:
Not certain of the relevance.  `model_utils.Choices` is a utility provided because working with `choices=` is unwieldly.  If Django introduces a better way to do it in future, then that will likely be something I'll migrate to, (assuming enums don't have too much magic), but for now, I want to solve my problem, and in doing so I believe I may be able to help others.

Not to mention `model_utils` supports [Django 1.11, and Python 2.7](https://github.com/jazzband/django-model-utils/blob/1c2992476f8d8a5fc627c32a1fd749cade81b346/tox.ini#L3).

PR #389, opened 2019-08-21
Fix excess number of queries

Natim:
Superseed #286


model_utils/managers.py
```diff
@@ -115,8 +115,12 @@ def annotate(self, *args, **kwargs):
         return qset
 
     def iterator(self):
-        # Maintained for Django 1.8 compatability
         iter = super(InheritanceQuerySetMixin, self).iterator()
+        if django.VERSION >= (1, 9):
+            for obj in iter:
```
Natim: What about returning `iter` directly there?

Natim:
The readme says django-model-utils supports Django 1.11 and 2.1+. So I guess this can be closed.

PR #392, opened 2019-09-16
Replace deprecated `django.utils.six` module with `six` library

unchris:
## Problem

I am preparing an application for Django 3 support using Django 3.0a1. I've noticed that `django-model-utils` still uses `django.utils.six` which is deprecated and removed in Django 3. 

## Solution

I replaced `django.utils.six` with `six` which can be fetched from pypi as usual. Added Python3/Django3 test target to `tox` (I'm not an expert at `tox` setup so please advise if I've screwed that up). 

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -5,7 +5,7 @@ import uuid
 from django.db import models
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.utils.encoding import python_2_unicode_compatible
+from six import python_2_unicode_compatible
 from django.utils.timezone import now
 
 DEFAULT_CHOICES_NAME = 'STATUS'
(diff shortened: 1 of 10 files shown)


unchris:
I have marked `Cover it with tests` because presumably we would just want to prevent regressions with the existing test coverage when changing a 3rd party dep.

unchris:
welp I have more work to do here than I thought

model_utils/fields.py
```diff
@@ -5,7 +5,7 @@
 from django.db import models
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.utils.encoding import python_2_unicode_compatible
+from six import python_2_unicode_compatible
```
auvipy: this will not needed as support for python 2 wil be dropped

unchris: I see, though I wasn't trying to drop python 2 support at the same time, just trying to solve the Django 3-alpha stuff.

I noticed you closed the PR. Is there something you would like me to do differently?

unchris: @auvipy ok, is there a different PR that's handling dropping Python2 or adding Django3 that I can help with?

PR #394, opened 2019-09-29
Drop Python 2 support.

hramezani:
## Problem

Drop Python 2 support.
https://github.com/jazzband/django-model-utils/issues/393

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -5,7 +5,6 @@ import uuid
 from django.db import models
 from django.conf import settings
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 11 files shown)


model_utils/fields.py
```diff
@@ -178,7 +177,6 @@ def get_excerpt(content):
     return '\n'.join(default_excerpt)
 
 
-@python_2_unicode_compatible
 class SplitText(object):
```
auvipy: remove object as well

hramezani:
> drop 2.7 from travis as well

Done!

PR #395, opened 2019-09-30
Add django 3.0 to the test matrix and drop six

mvbrn:
## Problem

Six won't be used in Django 3.0 so it will cause ImportError

## Solution

Drop six and add django 3.0 to the test matrix to ensure it is supported

## Commandments

- [x] Cover it with tests.


auvipy:
one last failure self = <django.db.models.query_utils.DeferredAttribute object at 0x7f6db0f84e80>

instance = <ModelWithCustomDescriptor: ModelWithCustomDescriptor object (2)>

    def _check_parent_chain(self, instance):

        """

        Check if the field value can be fetched from a parent field already

        loaded in the instance. This can be done if the to-be fetched

        field is a primary key field.

        """

        opts = instance._meta

>       link_field = opts.get_ancestor_link(self.field.model)

E       AttributeError: 'StringyDescriptor' object has no attribute 'model'

.tox/py37-django30/lib/python3.7/site-packages/django/db/models/query_utils.py:148: AttributeError

auvipy:
can you plz resolve the conflicts?

Surgo:
@mvbrn Any update?

tony:
@mvbrn Hm, conflicts need to be resolved on this, and I actually just bumped into django-model-utils when checking into Django 3.0 (it's released now)

Any status report on if you can get around to this?

In the event you are too busy, is it okay if other contributors fork this PR or work off it?

Natim:
`E AttributeError: 'StringyDescriptor' object has no attribute 'model'` with Python 3.8

Natim:
Ok something is probably wrong with this StringlyDescriptor mock.

Natim:
@lucaswiman since you wrote the test, do you have any insights about how to fix that?

auvipy:
can not load travis

Natim:
I am so sorry for the noise, since it is @mvbrn master branch I cannot push-force in it so I cannot merge my commit or squash anything...

auvipy:
OK 4 test failing

Natim:
Apparently the code coverage report decreasing is because we removed lines not test coverage.

Natim:
> OK 4 test failing

They are now fixed.

tests/models.py
```diff
@@ -373,26 +373,27 @@ class StringyDescriptor:
     """
     Descriptor that returns a string version of the underlying integer value.
     """
-
     def __init__(self, name):
-        self.name = name
```
Natim: I would revert that change.

CHANGES.rst
```diff
@@ -423,4 +426,4 @@ CHANGES
 0.3.0
 -----
 
-* Added ``QueryManager``
+- Added ``QueryManager``
```
Natim: This is up to you, I can also use `* ` everywhere. Just a matter of taste but I feel that we shouldn't mix them.

Natim:
Thank you @auvipy for the quick review!

PR #398, opened 2019-11-10
Modernize Python syntax, add Python 3.8

ad-m:


.travis.yml
```diff
@@ -1,10 +1,12 @@
-sudo: True
+version: ~> 1.0
 dist: xenial
```
auvipy: make it bionic

.travis.yml
```diff
@@ -1,10 +1,12 @@
...
+os: linux
 language: python
 cache: pip
 python:
-- 3.7
 - 3.6
+- 3.7
+- 3.8-dev
```
auvipy: this should 3.8 only

.travis.yml
ad-m: Done.

.travis.yml
ad-m: Done. I was not aware that TravisCI added it already.

auvipy:
all the travis builds are failing

ad-m:
@auvipy : fixed.

PR #410, opened 2020-01-03
Changed test case to reflect my setup better

RaddishIoW:
I've changed the test case to reflect better the setup in my code.

I have been unable to test this myself, as I don't have postgresql installed on this dev machine.

### DIFF
--- a/tests/models.py
+++ b/tests/models.py
@@ -369,8 +369,13 @@ class StatusFieldChoicesName(models.Model):
 
 
 class StatusFieldChoicesDisplay(models.Model):
-    NAMED_STATUS = (("UNCB", _("Unconfirmed Booking")), ("CONB", _("Confirmed Booking")))
-    status = StatusField(choices_name='NAMED_STATUS')
+    UNCONFIRMED_BOOKING = 'UNCB'
+    CONFIRMED_BOOKING = 'CONB'
+    JOB_STATUSES = ((UNCONFIRMED_BOOKING, _("Unconfirmed Booking")),
+                    (CONFIRMED_BOOKING, _("Confirmed Booking"))
+                    )
+    STATUS = JOB_STATUSES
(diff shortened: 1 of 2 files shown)


RaddishIoW:
Ah! Forgot to check the assert, but it still looks like it's not recreating the error I'm seeing...

tests/models.py
```diff
@@ -369,8 +369,13 @@ class StatusFieldChoicesName(models.Model):
...
 class StatusFieldChoicesDisplay(models.Model):
-    NAMED_STATUS = (("UNCB", _("Unconfirmed Booking")), ("CONB", _("Confirmed Booking")))
-    status = StatusField(choices_name='NAMED_STATUS')
+    UNCONFIRMED_BOOKING = 'UNCB'
+    CONFIRMED_BOOKING = 'CONB'
+    JOB_STATUSES = ((UNCONFIRMED_BOOKING, _("Unconfirmed Booking")),
+                    (CONFIRMED_BOOKING, _("Confirmed Booking"))
+                    )
+    STATUS = JOB_STATUSES
+    status = StatusField(default=CONFIRMED_BOOKING)
```
auvipy: any migration change needed in test?

RaddishIoW: Good point. I'm not sure, to be completely honest with you. I am interested if the type of quotes used for the definitions matters at all? i.e. I'm using `CONFIRMED_BOOKING = 'CONB'` and not `CONFIRMED_BOOKING = "CONB"`, if that might have an effect?
I'm also using sqlite in testing locally, but postgresql in production.

RaddishIoW:
There - that last commit has pinned down the issue I'm seeing. So it's something to do with the fact I'm using StatusModel as my model's superclass, then overriding the status field...

Natim:
Superseeded by #432

PR #412, opened 2020-01-21
Overriding TimeStampedModel's save() method

bugraaydin:
to update modified field automatically when update_fields argument is used but modified was not given as a parameter.

## Problem

When the save method was called with the 'update_fields' argument, the 'modified' field was not automatically updated when it was forgotten to be given as a parameter. This may lead to future confusion since developers may not be aware of this functionality, and it is more logical to expect this field to be updated even if it is not provided inside the 'update_fields'

## Solution

I've overrided the save method of the TimeStampedModel in order to support this functionality.

## Commandments

Unit tests on our product Plentific seems to work fine and as expected.

### DIFF
--- a/model_utils/models.py
+++ b/model_utils/models.py
@@ -35,7 +35,7 @@ class TimeStampedModel(models.Model):
         a parameter to the update field argument.
         """
         if 'update_fields' in kwargs and 'modified' not in kwargs['update_fields']:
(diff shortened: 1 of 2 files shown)


bugraaydin:
@auvipy will add tests and improve it within this week after finishing my other tickets.

PR #421, opened 2020-03-17
Document Django 3.0 enumeration types by 'Choices'

adamchainz:
Adding docs so others can discover enumeration types. Could also consider deprecating, at least when minimum supported Django is 3.0.

### DIFF
--- a/docs/utilities.rst
+++ b/docs/utilities.rst
@@ -4,8 +4,13 @@ Miscellaneous Utilities
 
 .. _Choices:
 
-Choices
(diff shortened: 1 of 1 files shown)


auvipy:
thanks!

adamchainz:
What about deprecating `Choices` at some point? Should I make an issue?

auvipy:
I believe yes! that should be deprecated eventually

adamchainz:
Made #422

Issue #399, opened 2019-11-13
StatusField is not using get_status_display since Django 2.2.7


RaddishIoW:
## Problem

Since updating Django to 2.2.7, get_status_display() doesn't return the human readable string.

## Environment

- Django Model Utils version: 3.2.0
- Django version: 2.2.7
- Python version: 3.6
- Other libraries used, if any:

## Code examples
```
UNCONFIRMED_BOOKING = 'UNCB'
CONFIRMED_BOOKING = 'CONB'
ASSINGNED = 'ASSD'
ACTIVE = 'ACTV'
COMPLETED = 'COMP'
CANCELLED = 'CANC'

JOB_STATUS = (
    (UNCONFIRMED_BOOKING, _("Unconfirmed Booking")),
    (CONFIRMED_BOOKING, _("Confirmed Booking")),
    (ASSINGNED, _("Assigned")),
    (ACTIVE, _("Active")),
    (COMPLETED, _("Completed")),
    (CANCELLED, _("Cancelled"))
)

class Job(StatusModel):
    '''
    Model for jobs
    '''

    STATUS = statuses.JOB_STATUS
    status = StatusField(db_index=True)
```
In template:
```
<p><b>Current status:</b> {{job.get_status_display}} </p>
```
For a status of CONFIRMED_BOOKING displays "CONB" instead of "Confirmed Booking"

Natim:
Hello @RaddishIoW can you try that this issue is still present with Django 2.2.8 and Django 3.0?

RaddishIoW:
Just updated to 2.2.8 and I'm still seeing it.

I'll try with 3.0, if I can get my project to run with it!

RaddishIoW:
Some of my other dependancies don't yet work with Django 3.0 (django-registration-redux, among others), so I can't test against 3.0 unless I write a special test case project.

RaddishIoW:
Could do with updating to 2.2.8 ASAP, as there's a security warning against 2.2.5... Is there a workaround I could implement? How can I get the display text of the status in a template?

Natim:
@RaddishIoW could you start a PR with a failing test for your use case?

Natim:
I looked at the tests, what about using:

```
JOB_STATUS = (
    (UNCONFIRMED_BOOKING, "Unconfirmed Booking", _("Unconfirmed Booking")),
    (CONFIRMED_BOOKING, "Confirmed Booking", _("Confirmed Booking")),
[...]

PR #432, opened 2020-06-12
Bug/399/status field get status display

Natim:
Refs #399


Natim:
Refs #399

foarsitter:
https://github.com/jazzband/django-model-utils/issues/399 seems to be resolved so this PR seems no longer applicable to me.

Issue #11, opened 2012-04-13
Inheritance for foreign keys


thibault:
Hi,

I cannot find a way to use the wonderful (it saves my life) InheritanceManager through a ForeignKey. Here's some code, it will be clearer:

```
class Product(Model):
    objects = InheritanceManager()

class Shirt(Product):
    …

class Jacket(Product):
    …

class Box(Model):
    product = ForeignKey('Product')

boxes = Box.objects.all()
for box in boxes:
    box.product # How to get shirts and jackets here?
```

Is there any way to do this? Cannot find it in the doc. If not, that would be a great feature to add.

Regards,
Thibault

AlexCid:
Hey ! I was looking for the solution of this problem and found this.I was wondering, did you actually fix this issue ? The proposed feature addition didn't seem to have made it into the master branch...

carljm:
@AlexCid No, nobody has written any code to implement the proposed feature. If they had and it had been merged, this issue would be closed.

AlexCid:
too bad :( CrazyCasta's message seemed to imply that the code was available but it seems i misunderstood him. Thanks for your rapid answer anyways !

AndrewHows:
I'm still beating my head against the ForeignKey field, trying to grok it, but...

Would it be possible to create a ForeignKey subclass that overrides however the hell it fetches its value, so that it goes through an InheritanceManager and calls select_subclasses()?

carljm:
@AndrewHows I'd think that would be possible, yeah. That's actually not a bad API for this; placing the decision on the FK side makes some sense.

AndrewHows:
Ok. If I get it working, I'll send a pull request

rrauenza:
I started trying to pick this apart as well (the ForeignKey idea) ... I started by using the trace class in Python:

```
In [11]: tracer.run('m.foreignkeyfield') 
 --- modulename: related_descriptors, funcname: __get__
related_descriptors.py(153):         if instance is None:
related_descriptors.py(159):         try:
related_descriptors.py(160):             rel_obj = getattr(instance, self.cache_name)
related_descriptors.py(161):         except AttributeError:
related_descriptors.py(162):             val = self.field.get_local_related_value(instance)
 --- modulename: related, funcname: get_local_related_value
related.py(597):         return self.get_instance_value_for_fields(instance, self.local_related_fields)
 --- modulename: related, funcname: local_related_fields
[...]

PR #434, opened 2020-07-06
managers: avoid querying too much tables when not needed

Hyask:
## Problem

When calling `select_related()` with an empty list of arguments (1), Django will
try to prefetch some data by doing some first level joints with the related
classes.
This can lead to obvious negative performance impact, but this also breaks some
workarounds for having inheritance for foreign keys (2), as those solutions rely
on lazy evaluation of the related object.

(1): https://github.com/django/django/blob/a4e6030904df63b3f10aa0729b86dc6942b0458e/django/db/models/query.py#L1051
    Only passing an explicit `None` to `select_related` will disable the magic.
(2): https://github.com/jazzband/django-model-utils/issues/11

As examples, here are the generated SQL requests in
InheritanceManagerRelatedTests.test_get_method_with_select_subclasses_check_for_useless_join:

  * without this fix, without adding `.select_related(None)`
```sql
SELECT
    "tests_inheritancemanagertestparent"."id",
    "tests_inheritancemanagertestparent"."non_related_field_using_descriptor",
    "tests_inheritancemanagertestparent"."related_id",
    "tests_inheritancemanagertestparent"."normal_field",
    "tests_inheritancemanagertestparent"."related_self_id",
    "tests_inheritancemanagertestchild4"."other_onetoone_id",
    "tests_inheritancemanagertestchild4"."parent_ptr_id", T3."id",
    T3."non_related_field_using_descriptor", T3."related_id", T3."normal_field",
    T3."related_self_id"
FROM
    "tests_inheritancemanagertestchild4"
INNER JOIN
    "tests_inheritancemanagertestparent" ON
    ("tests_inheritancemanagertestchild4"."parent_ptr_id" = "tests_inheritancemanagertestparent"."id")
INNER JOIN
    "tests_inheritancemanagertestparent" T3 ON
    ("tests_inheritancemanagertestchild4"."other_onetoone_id" = T3."id")
WHERE
    "tests_inheritancemanagertestchild4"."parent_ptr_id" = 191
```

  * with either the fix, or by adding `.select_related(None)` after `.select_subclasses()`
```sql
SELECT
    "tests_inheritancemanagertestparent"."id",
    "tests_inheritancemanagertestparent"."non_related_field_using_descriptor",
    "tests_inheritancemanagertestparent"."related_id",
    "tests_inheritancemanagertestparent"."normal_field",
    "tests_inheritancemanagertestparent"."related_self_id",
    "tests_inheritancemanagertestchild4"."other_onetoone_id",
    "tests_inheritancemanagertestchild4"."parent_ptr_id"
FROM
    "tests_inheritancemanagertestchild4"
INNER JOIN
    "tests_inheritancemanagertestparent" ON
    ("tests_inheritancemanagertestchild4"."parent_ptr_id" = "tests_inheritancemanagertestparent"."id")
WHERE
    "tests_inheritancemanagertestchild4"."parent_ptr_id" = 191

```

## Solution

Add a check so that the `InheritanceQuerySetMixin` doesn't force Django to query unneeded tables.

## Commandments

- [x] Cover it with tests.
- [ ] ~~Update documentation (if relevant).~~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ CHANGES
 
 4.0.1 (unreleased)
 ------------------
+- Update InheritanceQuerySetMixin to avoid querying too much tables
 - TimeStampedModel now automatically adds 'modified' field as an update_fields
   parameter even if it is forgotten while using save()
 - `FieldTracker` now marks fields as not changed after `refresh_from_db`
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -75,7 +75,10 @@ class InheritanceQuerySetMixin:
 
         # workaround https://code.djangoproject.com/ticket/16855
         previous_select_related = self.query.select_related
-        new_qs = self.select_related(*subclasses)
+        if subclasses:
+            new_qs = self.select_related(*subclasses)
+        else:
+            new_qs = self
         previous_is_dict = isinstance(previous_select_related, dict)
         new_is_dict = isinstance(new_qs.query.select_related, dict)
         if previous_is_dict and new_is_dict:
--- a/tests/test_managers/test_inheritance_manager.py
+++ b/tests/test_managers/test_inheritance_manager.py
@@ -487,6 +487,14 @@ class InheritanceManagerRelatedTests(InheritanceManagerTests):
                 id=self.child1.id),
             self.child1)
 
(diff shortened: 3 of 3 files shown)


Hyask:
@auvipy do I have something to do to merge this, or is it on your side?

Issue #435, opened 2020-07-27
Incorrect FieldTracker behaviour if primary key is not AutoField


MaximZemskov:
## Problem

In cases when primary key is not `AutoField`(e.g. `UUIDField`) `FieldTracker` has different behaviour

## Environment

- Django Model Utils version: 4.0.0
- Django version: 2.2.13
- Python version: 3.8

### Correct behaviour:

Model
```python
class Test(models.Model):
    t_field = models.CharField(max_length=120)

    tracker = FieldTracker(fields=('t_field',))
```

Input:
```python
t = Test(t_field='test')
print(t.tracker.changed())
```

Output:
`>>>  {'t_field': None}`

### Incorrect behaviour

Model
```python
class Test(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    t_field = models.CharField(max_length=120)

    tracker = FieldTracker(fields=('t_field',))
```

Input:
```python
t = Test(t_field='test')
print(t.tracker.changed())
```

Output:
`>>>  {}`

MRigal:
As seen in the failing tests in the PR, this is not working with defferred fields, so the fix isn't so easy. Anyone can try to fix it in the open PR

PR #436, opened 2020-07-27
#435 Fix instance creation state check

MaximZemskov:
## Problem

FieldTracker has incorrect behaviour if PK field isn't a `AutoFIeld`. Link to issue https://github.com/jazzband/django-model-utils/issues/435.

## Solution

Django has a [`ModelState`](https://github.com/django/django/blob/b7b7df5fbcf44e6598396905136cab5a19e9faff/django/db/models/base.py#L393) object to determine if objects is in creation state or already exists in db. Replacing `if not self.instance.pk is None` with `if self.instance._state.adding` solve the problem.

## Commandments

- [ ] Cover it with tests.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -49,6 +49,7 @@
 | Lucas Wiman <lucaswiman@counsyl.com>
 | Martey Dodoo <martey@mobolic.com>
 | Matthew Schinckel <matt@schinckel.net>
+| Maxim Zemskov <m.zemskov1@gmail.com>
 | Michael van Tellingen <m.vantellingen@lukkien.com>
 | Mike Bryant <m@ocado.com>
 | Mikhail Silonov <m.silonov@corp.mail.ru>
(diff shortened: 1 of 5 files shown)


MaximZemskov:
@auvipy I fixed one of two failed tests. Trying to fix last test but can't figure out what is the problem

martey:
@auvipy I'm not sure why a review by me was requested. Was this an error, or do you want me to help with something?

auvipy:
GitHub shows you recently modified the files that's why I requested a review

martey:
Looking at my previous contributions to this project, the only file I have edited that is in this pull request is `AUTHORS.rst`. Since this pull request does not contain substantial changes to that file, I will refrain from providing a review.

MaximZemskov:
Looks like at the moment of execution `set_saved_fields` method `self.instance._state.adding` is always `True`. It becomes `False` inside django's `Model.from_db` method

MaximZemskov:
Found the problem.
`initialize_tracker` called on model's `post_init` signal and `_state` at this moment contains default values
https://docs.djangoproject.com/en/3.0/ref/signals/#post-init

> Note
>instance._state isn’t set before sending the post_init signal, so _state attributes always have their default values. For example, _state.db is None.

PR #438, opened 2020-08-12
Add available_objects soft deletable model manager

craiga:
## Problem

As described in #364

## Solution

 * Adds `.available_objects` manager to soft deletable models.
 * Keeps existing `.objects` manager, but emits a `DeprecationWarning` for each use of that manager.

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/tests/test_models/test_softdeletable_model.py
+++ b/tests/test_models/test_softdeletable_model.py
@@ -6,45 +6,48 @@ from tests.models import SoftDeletable
 
 class SoftDeletableModelTests(TestCase):
     def test_can_only_see_not_removed_entries(self):
-        SoftDeletable.objects.create(name='a', is_removed=True)
-        SoftDeletable.objects.create(name='b', is_removed=False)
+        SoftDeletable.available_objects.create(name='a', is_removed=True)
+        SoftDeletable.available_objects.create(name='b', is_removed=False)
 
-        queryset = SoftDeletable.objects.all()
+        queryset = SoftDeletable.available_objects.all()
 
         self.assertEqual(queryset.count(), 1)
         self.assertEqual(queryset[0].name, 'b')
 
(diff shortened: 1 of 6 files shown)


craiga:
Still needs documentation. Just hoping for feedback on this approach for now.

jonathan-s:
This approach looks sane to me. Perhaps the link should be to the documentation instead for this? With regards to emitting too many warnings, I believe there is a way to silence warnings which could be mentioned in the documentation. If for nothing else this will bring people the notice of how things will work in the future and change things accordingly.

tests/test_models/test_softdeletable_model.py
```diff
@@ -6,45 +6,49 @@
...
         self.assertEqual(SoftDeletable.all_objects.count(), 0)
 
     def test_instance_purge_no_connection(self):
-        instance = SoftDeletable.objects.create(name='a')
+        instance = SoftDeletable.available_objects.create(name='a')
 
         self.assertRaises(ConnectionDoesNotExist, instance.delete,
                           using='other', soft=False)
+
+    def test_can_only_see_not_removed_entries(self):
```
craiga, line 52: Test named incorrectly.

robypomoni (approved):

rjekker:
One thing is not clear for me:

does this mean that in the future, `objects` will again be the default Manager that returns everything?

What is currently the recommended way to get to deleted items when using SoftDeletableModel?

craiga:
@rjekker Yep, that's right—`objects` will be a plain old simple manager.

I might be wrong, but there is no recommended way to get deleted items at the moment. That's part of the underlying issue this deprecation is trying to solve.

tadamcz:
Thanks for moving this along, @craiga! I added some clarification to the docs (https://github.com/jazzband/django-model-utils/pull/589) as I was initially confused.

PR #446, opened 2020-10-01
Issues with TimeStamped model's 'modified' and 'update_fields'

arseniy-panfilov:
## Problem

There are a few issues with current implementation of `TimeStamped.save()` method related to how it handles the `update_fields` argument.
Here's how current implementation diverges from [Django docs](https://docs.djangoproject.com/en/3.1/ref/models/instances/#specifying-which-fields-to-save):
1. It expects `update_fields` to be a `list`, while it can be any iterable:
> The update_fields argument can be any iterable containing strings.
2. When passed an empty list, it will add `modified` to it thus triggering an unexpected save:
> An empty update_fields iterable will skip the save.
3. Explicitly passing `update_fields=None` causes `TypeError`, while it's a valid value:
> A value of None will perform an update on all fields.

## Solution

This PR adds a few checks to `TimeStamped.save` method to ensure that:
1. Any iterable is supported - by converting `update_fields` it to a `set` before trying to add the `'modified'` field to it. `set` is chosen so that we can avoid explicit `'modified' in updated_fields` check and rely on uniqueness of its elements instead;
2. `'modified'` not added to an empty `updated_fields`;
3. `None` value passed to the `super` method as it is.

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/model_utils/models.py
+++ b/model_utils/models.py
@@ -34,8 +34,12 @@ class TimeStampedModel(models.Model):
         modified field is updated even if it is not given as
         a parameter to the update field argument.
         """
-        if 'update_fields' in kwargs and 'modified' not in kwargs['update_fields']:
-            kwargs['update_fields'] += ['modified']
+        update_fields = kwargs.get('update_fields', None)
+        if update_fields is not None:
+            update_fields = set(update_fields)
+            if update_fields:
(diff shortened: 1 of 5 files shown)


auvipy (approved):

sebatyler:
Is there any plan to release a package including this PR?

Issue #400, opened 2019-11-21
Django 3.0 support


domdinicola:
Add support for Django 3.0

good that python 2.x support has been already dropped https://github.com/jazzband/django-model-utils/issues/393

spielmannj:
One thing of note is that this means that the `django.utils.six` module was dropped.

ephes:
Just using the master instead of the package worked for me:

git+git://github.com/jazzband/django-model-utils.git#egg=django-model-utils

domdinicola:
@ephes I would rater wait for a package release

ephes:
> @ephes I would rater wait for a package release

Yes, for important projects I would also wait. But I tried this for a small private project and was surprised how many packages break on django 3 because those old python 2 stuff.

domdinicola:
it's cause python 2 is still supported until the end of the year.. people were mostly busy upgrading to py3 (using six)

bixbyr:
I guess other people already discovered that the package is broken in django 3.0. 

Is there a timeline for releasing the next version? I'd like to stick to using a package instead of just pulling from master :(

Natim:
I just found out that master was already supporting Django 3.0 but I am also interested in a release of this current version even if there are works on the way to remove six (#395)

Natim:
#395 has just been merged.

@auvipy can we make a release or do you see any mandatory work to be made first?

smithdc1:
@hramezani are you able to push a release? Seems a shame to have so many commits on master since the last release?

hramezani:
@smithdc1 No. Maybe @jezdez can help.

smithdc1:
According to https://jazzband.co/projects/django-model-utils @charettes (:wave: ) is a project lead?

jezdez:
Yeah, let's prepare a release and wait if @charettes responds first.

charettes:
Should be released in a few hours :)

charettes:
Should be released now https://pypi.org/project/django-model-utils/. Thanks to everyone involve and sorry for the delay here.

PR #462, opened 2020-11-29
Migrate to GitHub Actions

jezdez:
Travis CI has a new pricing model which places limits on open source.

* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
* https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Many projects are moving to GitHub Actions instead, including Jazzband projects:

* https://github.com/orgs/jazzband/projects/1
* https://github.com/jazzband-roadies/help/issues/206

This is based on https://github.com/jazzband/contextlib2/pull/26.

TODO:

* [x] @jezdez to add `JAZZBAND_RELEASE_KEY` to the repo secrets.

PR #463, opened 2020-11-29
Release 4.1.0

charettes:
Refs #400

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 CHANGES
 =======
 
-4.0.1 (unreleased)
+4.1.0 (2020-11-29)
 ------------------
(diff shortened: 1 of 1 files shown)


jezdez:
Could we have #462 in there as well?

charettes:
Sure thing, :eyes:

jezdez:
@charettes Once #462 is merged the version number is derived from the Git tag, not needed to be updated, using [setuptools_scm](https://pypi.org/project/setuptools-scm/). The rest of the release process stays the same.

illing2005:
This should have been a new major version (5.0.0) because it has some breaking changes (like dropping django 2.1)
Took me quite some time to figure out why my CI/CD pipeline wasn't working anymore :wink:

Issue #465, opened 2020-12-01
Inheriting model from TimeStampModel and StatusModel results in TypeError


LHolst:
## Problem
Inheriting custom model class from TimeStampModel and StatusModel results in TypeError

StatusModel.save updates the 'update_fields` kwargs like
```
kwargs['update_fields'] += ['status_changed']
```

maybe related to https://github.com/jazzband/django-model-utils/issues/464
## Environment

- Django Model Utils version: 4.1.0
- Django version: 2.2
- Python version: 3.6
- Other libraries used, if any:

## Code examples

```py
class MyModel(TimeStampModel, StatusModel):
    .....


#later calling 
my_model.save(update_fields=('fieldName',))
 
#results in 
TypeError: unsupported operand type(s) for +=: 'set' and 'list'
```

LHolst:
should StatusModel.save convert the `kwargs['update_fields']` too ?

bugraaydin:
Hello @LHolst, yes exactly, a similar fix was present in `TimeStampedModel`. I forgot to cover that case and thought the actual fix handled `StatusModel` as well, do you have capacity to implement a similar manner like the `TimeStampedModel`'s implementation

hramezani:
This problem introduced in https://github.com/jazzband/django-model-utils/pull/446 because we save a `set` in `kwargs['update_fields']` at `TimeStampedModel.save`  that is inconsistent with `StatusModel.save`(it saves list).

I created a [PR](https://github.com/jazzband/django-model-utils/pull/466) and changed both save method to save `list` in `kwargs['update_fields']`

bugraaydin:
@hramezani sorry about me introducing this problem, i have noticed this missing functionality in our codebase and wanted to contribute back after implementing it for my code as well; was first months of my software engineering experience and implementation was not covering all the possible cases.

hramezani:
@bugraaydin Actually the changes that we made in https://github.com/jazzband/django-model-utils/pull/446 cause the problem I think

bugraaydin:
> @bugraaydin Actually the changes that we made in #446 cause the problem I think

I have introduced the problematic code in #412 and #441, the PR you have linked was an attempt to fix it I suppose but it introduces another problem, inconsistency between kwargs types if the model is using both classes

PR #466, opened 2020-12-01
Fixes save error when model inherit from both TimeStampModel and StatusModel.

hramezani:
## Problem

Fixed https://github.com/jazzband/django-model-utils/issues/465

## Solution

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/model_utils/models.py
+++ b/model_utils/models.py
@@ -32,10 +32,8 @@ class TimeStampedModel(models.Model):
         a parameter to the update field argument.
         """
         update_fields = kwargs.get('update_fields', None)
-        if update_fields is not None:
-            update_fields = set(update_fields)
-            if update_fields:
-                kwargs['update_fields'] = update_fields.union({'modified'})
+        if update_fields:
+            kwargs['update_fields'] = set(update_fields).union({'modified'})
 
         super().save(*args, **kwargs)
 
@@ -74,12 +72,10 @@ class StatusModel(models.Model):
         status_changed field is updated even if it is not given as
         a parameter to the update field argument.
         """
-        if (
-            'update_fields' in kwargs
-            and 'status' in kwargs['update_fields']
-            and 'status_changed' not in kwargs['update_fields']
-        ):
-            kwargs['update_fields'] += ['status_changed']
+        update_fields = kwargs.get('update_fields', None)
+        if update_fields and 'status' in update_fields:
+            kwargs['update_fields'] = set(update_fields).union({'status_changed'})
+
         super().save(*args, **kwargs)
 
     class Meta:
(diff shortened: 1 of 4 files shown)


hramezani:
@arseniy-panfilov Could you please take a look at this patch?

model_utils/models.py
```diff
@@ -32,10 +32,8 @@ def save(self, *args, **kwargs):
         a parameter to the update field argument.
         """
         update_fields = kwargs.get('update_fields', None)
-        if update_fields is not None:
-            update_fields = set(update_fields)
-            if update_fields:
-                kwargs['update_fields'] = update_fields.union({'modified'})
+        if update_fields and 'modified' not in update_fields:
+            kwargs['update_fields'] = list(kwargs['update_fields']) + ['modified']
```
charettes, line 36: I think the previous approach was fine.

```suggestion
        if update_fields:
            kwargs['update_fields'] = set(update_fields).union({'modified'})
```

model_utils/models.py
```diff
@@ -74,12 +72,14 @@ def save(self, *args, **kwargs):
...
         if (
-            'update_fields' in kwargs
-            and 'status' in kwargs['update_fields']
-            and 'status_changed' not in kwargs['update_fields']
+            update_fields
+            and 'status' in update_fields
+            and 'status_changed' not in update_fields
         ):
-            kwargs['update_fields'] += ['status_changed']
+            kwargs['update_fields'] = list(kwargs['update_fields']) + ['status_changed']
```
charettes, line 81: I think we should keep using `set` as above.

```suggestion
            kwargs['update_fields'] = set(update_fields).union({'status_changed'})
```

bugraaydin:
```
        update_fields = kwargs.get('update_fields', None)
        if update_fields is not None:
            update_fields = set(update_fields)
            if (
                update_fields
                and 'status' in update_fields
            )
        ):
            kwargs['update_fields'] = update_fields.union({'status_changed'})
```

changing the code here to above would do the trick i suppose

model_utils/models.py
hramezani: Thanks @charettes  I changed it to set approach

charettes (approved):
Thanks for the changes @hramezani 🙇

hramezani:
Thanks for the review and suggestion @charettes.
Should we prepare a bug fix release?

charettes:
@hramezani yeah I think it's worth doing. I can tag and release once this is merged if you can adjust the `CHANGES` with these adjustments.

bugraaydin:
> Thanks for the review and suggestion @charettes.
> Should we prepare a bug fix release?

sorry about causing the headache here again, thanks for the quick fix

hramezani:
@charettes I added an entry to `CHANGES`

PR #494, opened 2021-08-29
#491 Context manager for resetting fields

### LINKED ISSUE #491: Undocument breaking changes in FieldTracker in Version 4.1
## Problem

The Release 4.1, after moving the patch from the `save()` to the `save_base()` method in #404 (@tumb1er FYI) introduced a serioulsy breaking change that was neither announced nor documented (and was obviously not tested brefore). It would have deserved a 5.x update IMHO.

Since the values of the tracker are restored after the call to the `save_base()` method, any check/assumption made on the tracker inside the `save()` Method but **after** the `super().save()` call fails, as the tracker is then "new/empty". This is especially problematic when tracking changes depending on related objects, when working with nested serializers per example.

## Environment

- Django Model Utils version: <4.1 VS 4.1+

## Code examples

```
class MyModel(Model):
        name = models.CharField(max_length=64)
        tracker = FieldTracker()

        def save(self, *args, **kwargs):
            if self.tracker.changed('name'):
                 print("True")
            super().save(*args, **kwargs)
            if self.tracker.changed('name'):
                 print("True")
```
Just calling `MyModel.create()` would result in 2 prints in version <4.1 and in only one in version 4.1+ (the second check will fail as the tracker is resetted on the `save_base()` call inside the `super().save() call)

## Discussion / proposal

@auvipy A work-around is to test before the call to `save()` and then use the variable afterwards for such cases. 

But I honestly find it quite a dramatic regression for the IMHO limited benefit of being able to modify `update_fields` in the `save()` Method.

I would seriously prefer to restore the previous behaviour of patching the `save()` Method, but still add a functionality to support overriding `update_fields`: I would prefer to add a new kwarg `_always_update` to `FieldTracker` where you could define the list of fields that you always want to update. I think also in term of design and readbility it would be a better feature. @tumb1er what do you think about the suggestion? If you validate it, I could draft a PR for this

tumb1er:
## Problem

* There are breaking changes in `FieldTracker` state reset behavior described in #491
* Some people need old behavior, another new one

## Solution

This PR introduces `FieldsContext` context manager which allows to postpone state reset till outermost context exit

## Commandments

- [x] Cover it with tests.

Closes #491

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,13 @@ Unreleased
 4.1.1 (2020-12-01)
 ------------------
 
+- Applied `isort` to codebase (Refs GH-#402)
+- Fix `TypeError` in save when model inherits from both TimeStampModel
+  and StatusModel. (Fixes GH-465)
+
+4.1.0 (2020-11-29)
+------------------
+
 **Breaking changes:**
 - `FieldTracker` now marks fields as not changed after `refresh_from_db`
   respecting `fields` argument (GH-#404)
@@ -22,18 +29,9 @@ Unreleased
   methods (GH-#404)
 
 **Other changes:**
-- Applied `isort` to codebase (Refs GH-#402)
-- Fix `TypeError` in save when model inherits from both TimeStampModel
-  and StatusModel. (Fixes GH-465)
-
-4.1.0 (2020-11-29)
-------------------
 - Update InheritanceQuerySetMixin to avoid querying too much tables
 - TimeStampedModel now automatically adds 'modified' field as an update_fields
   parameter even if it is forgotten while using save()
-- `FieldTracker` now marks fields as not changed after `refresh_from_db`
-- `FieldTracker` now respects `update_fields` changed in overridden `save()`
-  method
 - Replace ugettext_lazy with gettext_lazy to satisfy Django deprecation warning
 - Add available_objects manager to SoftDeletableModel and add deprecation
   warning to objects manager.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 - Drop support for `Django 3.0`
 
 - Added urlsafe token field.
+- Introduce context manager for FieldTracker state reset (GH-#491)
 
 4.1.1 (2020-12-01)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,17 @@ Unreleased
 
 4.1.1 (2020-12-01)
 ------------------
+
+**Breaking changes:**
+- `FieldTracker` now marks fields as not changed after `refresh_from_db`
+  respecting `fields` argument (GH-#404)
+- `FieldTracker` now respects `update_fields` changed in overridden `save()`
+  method (GH-#404)
+- `FieldTracker` now resets states after `pre_save()` and not anymore `save()`
+  signals, possibly altering the behaviour of overridden `save()`
+  methods (GH-#404)
+
+**Other changes:**
 - Applied `isort` to codebase (Refs GH-#402)
 - Fix `TypeError` in save when model inherits from both TimeStampModel
   and StatusModel. (Fixes GH-465)
--- a/docs/utilities.rst
+++ b/docs/utilities.rst
@@ -346,3 +346,75 @@ This is how ``FieldTracker`` tracks field changes on ``instance.save`` call.
 8. ``instance.refresh_from_db()`` call causes initial state reset like for
    ``save_base()``.
 
+When FieldTracker resets fields state
+-------------------------------------
+
+By the definition:
+
+.. NOTE::
+    * Field value *is changed* if it differs from current database value.
+    * Field value *was changed* if value has changed in database and field state didn't reset.
+
+.. code-block:: python
+
+    instance = Tracked.objects.get(pk=1)
+    # name not changed
+    instance.name += '_changed'
+    # name is changed
+    instance.save()
+    # name is not changed again
+
+Current implementation resets fields state after ``post_save`` signals emitting. This is convenient for "outer" code
+like in example above, but does not help when model ``save`` method is overridden.
+
+.. code-block:: python
+
+    class MyModel(models.Model)
+        name = models.CharField(max_length=64)
+        tracker = FieldsTracker()
+
+        def save(self):  # erroneous implementation
+            self.name = self.name.replace(' ', '_')
+            name_changed = self.tracker.has_changed('name')
+            super().save()
+            # changed state has been reset here, so we need to store previous state somewhere else
+            if name_changed:
+                do_something_about_it()
+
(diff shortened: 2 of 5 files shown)


docs/utilities.rst
```diff
@@ -346,3 +346,75 @@ This is how ``FieldTracker`` tracks field changes on ``instance.save`` call.
...
+            if name_changed:
+                do_something_about_it()
+
+``FieldTracker`` provides a context manager interface to postpone fields state reset in complicate situations.
+
+* Fields state resets after exiting from outer-most context
+* By default, all fields are reset, but field list can be provided
+* Fields are counted separately depending on field list passed to context managers
+* Tracker can be used as decorator
+* Different instances have it's own context state
```
MRigal, line 390: typo `have their own`

model_utils/tracker.py
```diff
@@ -86,11 +86,46 @@ def __delete__(self, obj):
         self.descriptor.__delete__(obj)
 
 
+class FieldsContext:
+    def __init__(self, tracker, *fields, state=None):
+        if state is None:
+            state = {}
```
MRigal, line 92: Or maybe a `defaultdict(0)` ?

model_utils/tracker.py
```diff
@@ -86,11 +86,46 @@ def __delete__(self, obj):
         self.descriptor.__delete__(obj)
 
 
+class FieldsContext:
+    def __init__(self, tracker, *fields, state=None):
```
MRigal, line 90: Maybe you could add a docstring to explain the reason for this second Context Manager. I know this is to handle several states and it is clear when looking at the whole PR now, but might be less obvious when looking at the class inside the code later on.

model_utils/tracker.py
```diff
@@ -86,11 +86,46 @@ def __delete__(self, obj):
...
+        self.context = FieldsContext(self, *self.fields)
+
+    def __enter__(self):
+        return self.context.__enter__()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return self.context.__exit__(exc_type, exc_val, exc_tb)
+
+    def __call__(self, *fields):
+        return FieldsContext(self, *fields, state=self.context.state)
```
MRigal, line 128: Isn't it possible to `return self.context` directly here?

model_utils/tracker.py
```diff
@@ -240,21 +289,17 @@ def _patch(self, model, method, fields_kwarg):
...
         @wraps(original)
         def inner(instance, *args, **kwargs):
-            ret = original(instance, *args, **kwargs)
             update_fields = kwargs.get(fields_kwarg)
-            if not update_fields and update_fields is not None:  # () or []
-                fields = update_fields
-            elif update_fields is None:
-                fields = None
+            if update_fields is None:
+                fields = self.fields
```
MRigal, line 294: Nice you took that improvement from my PR ;-)

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Unreleased
 - Drop support for `Django 3.0`
 
 - Added urlsafe token field.
+- Introduce context manager for FieldTracker state reset (GH-#491)
 
 4.1.1 (2020-12-01)
 ------------------
```
MRigal, line 13: Although it is has already been released, IMHO we could change to the release notes of 4.1.0 to have
```
**Breaking changes:**
- `FieldTracker` now marks fields as not changed after `refresh_from_db`
- `FieldTracker` now respects `update_fields` changed in overridden `save()`
  method
- `FieldTracker` now resets states after `pre_save()` and not anymore `save()` signals, possibly altering the behaviour of overriden `save()` methods

**Review by MRigal — CHANGES_REQUESTED** [review]
This is really brilliant @tumb1er !!! Sorry for having taken so long to make the review. I had given a first look the first day, but I wanted to read it a couple more time to be sure to get the whole picture.

I've added various suggestions, feel free to take the ones you like! :-)

IMHO this deserves a `5.0` release

File: CHANGES.rst (hunk 6 above)
Comment (tumb1er) [reply CHANGES.rst:13]: Great idea! Added this mark.

File: docs/utilities.rst (hunk 1 above)
Comment (tumb1er) [reply docs/utilities.rst:390]: Fixed

File: model_utils/tracker.py (hunk 3 above)
Comment (tumb1er) [reply model_utils/tracker.py:90]: Added a docstring.

File: model_utils/tracker.py (hunk 2 above)
Comment (tumb1er) [reply model_utils/tracker.py:92]: `state` may be passed as a parameter to `FieldsContext.__init__`, it will be not so obvious that there should be `defaultdict` instead of `dict`.

File: model_utils/tracker.py (hunk 4 above)
Comment (tumb1er) [reply model_utils/tracker.py:128]: `self.context` will not allow to pass custom fields list

```python
with self.tracker('status'):
    with self.tracker('error'):
        self.produce_error()
        self.handle_error()
    self.handle_status()
```

model_utils/tracker.py
tumb1er: Yep, it's much more simple to understand.

model_utils/tracker.py
MRigal: Sure, my bad!

model_utils/tracker.py
```diff
@@ -86,11 +86,65 @@ def __delete__(self, obj):
...
+
+    >>> with fields_context(obj.tracker, 'f1', state=state):
+    ...     with fields_context(obj.tracker, 'f1', 'f2', state=state):
+    ...         obj.do_something_useful()
+    ...     # f2 is reset after inner context exit
+    ...     obj.do_something_else()
+    ... # f1 is reset after outer context exit
+    >>>
+
+    * Note that fields are countedbe passing same state dict
```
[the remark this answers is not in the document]
tumb1er: Now I'm not also :) I guess I meant "counted by".

I'll fix this after another discussion https://github.com/jazzband/django-model-utils/pull/494#discussion_r724006411

model_utils/tracker.py
MRigal: Hi @tumb1er I'm not really sure this will reduce the complexity, so I guess you'd better leave it as it is. Maybe you could also here add a docstring that explains the architecture.

tumb1er: Complexity - bad, docstrings - good. I'll do it.

auvipy: is it possible to split this pr in smaller ones?

tumb1er: In this PR:
* Two classes for new mechanics were introduced for fields state tracking
* Some helper methods added that simplify using new mechanics in new code
* default monkey-patched methods refactored to use new mechanics

I can split it, but each one will be more or less useless without another one.

tumb1er:
@auvipy @MRigal still one typo and docstring are waiting to be fixes.

MRigal:
LGTM. @auvipy up to you to check another time this MR, but IMHO it can be merged and is a great addition.

auvipy (approved):

PR #495, opened 2021-08-29
Fix QuerySet._chain for django main (post-3.2)

tumb1er:
## Problem

Tests are failing agains django main branch (post-3.2) because of `QuerySet._chain` signature change, see https://github.com/django/django/pull/14755/files:

```
# was
def _chain(self, **kwargs): ...

# now
def _chain(self): ...
```

## Solution

Update additional fields manually

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -86,11 +86,14 @@ class InheritanceQuerySetMixin:
         return new_qs
 
     def _chain(self, **kwargs):
+        update = {}
         for name in ['subclasses', '_annotated']:
             if hasattr(self, name):
-                kwargs[name] = getattr(self, name)
+                update[name] = getattr(self, name)
 
-        return super()._chain(**kwargs)
+        chained = super()._chain(**kwargs)
+        chained.__dict__.update(update)
+        return chained
 
     def _clone(self, klass=None, setup=False, **kwargs):
         qs = super()._clone()


auvipy (changes requested):
this needs unit tests

tumb1er:
@auvipy  Without that change 36 tests are already failing on django main with:

```
TypeError: _chain() got an unexpected keyword argument 'subclasses'
```

What exactly should be tested?

tumb1er:
More interesting: if I just delete `_chain` method, no tests are failing. I think that's because `_chain` calls `_clone` internally, and that one does same thing.

model_utils/managers.py
```diff
@@ -86,11 +86,14 @@ def select_subclasses(self, *subclasses):
...
 
     def _chain(self, **kwargs):
+        update = {}
         for name in ['subclasses', '_annotated']:
             if hasattr(self, name):
-                kwargs[name] = getattr(self, name)
+                update[name] = getattr(self, name)
 
-        return super()._chain(**kwargs)
+        chained = super()._chain(**kwargs)
```
MRigal, line 94: Isn't it risky to still pass kwargs here? Sure it should be empty in Django 4+

auvipy (approved):

PR #499, opened 2021-10-07
Add support for Python 3.10

joshuadavidthomas:
## Problem

Python 3.10 was released this past Monday. 

This codebase had an incompatible line in `model_utils\fields.py`. It was importing `Callable` from `collections`, which the ability to do so was removed in 3.10.

One of the dependencies, pytest, was out of date as well.

## Solution

- change the import to `from collections.abc import Callable`
- update pytest to 6.2.5 which is 3.10 compatible
- add 3.10 to the GH action strategy matrix and py310 to tox.ini, with [passing tests](https://github.com/joshuadavidthomas/django-model-utils/actions)
- update the trove identifiers

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ Unreleased
 ----------
 - Add support for `Django 3.2`
 - Drop support for `Django 3.0`
-
+- Add support for `Python 3.10`
 - Added urlsafe token field.
 - Introduce context manager for FieldTracker state reset (GH-#491)
 


hramezani:
It seems Django 2.2 is not compatible with Python 3.10
https://github.com/django/django/pull/14942#issuecomment-934305693

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
-    py{36,37,38,39}-dj{22,31,32}
-    py{38,39}-djmain
+    py{36,37,38,39,310}-dj{22,31,32}
+    py{38,39,310}-djmain
```
MRigal, line 4: @hramezani is right, I've overseen that! Please change to 
```
py{36,37,38,39,310}-dj{22,31,32}
py310-dj{31,32}
```

joshuadavidthomas: Wouldn't it be
```python
py{36,37,38,39}-dj{22,31,32}
py310-dj{31,32}
```
to exclude Django 2.2 from Python 3.10? This may be a silly question, but I'm a bit of a noob to tox, so I'm still getting acclimated to the ini config syntax.

PR #513, opened 2021-12-07
Update django support

hramezani:
## Problem

* Remove Django 3.1 from Python 3.10 env in tox.

 * Confirm support for Django 4.0

* Add USE_TZ setting.

## Solution

## Commandments

- [ ] Cover it with tests.


setup.py
```diff
@@ -47,6 +47,7 @@ def long_desc(root_path):
         'Framework :: Django :: 2.2',
         'Framework :: Django :: 3.1',
         'Framework :: Django :: 3.2',
+        'Framework :: Django :: 4.0',
     ],
     zip_safe=False,
     tests_require=['Django>=2.2'],
```
adamchainz, line 53: Shouldn't `install_requires` also say `>=2.2` ? I don't think there's even a need to put `django` in `tests_require` if it's in `install_requires`.

tox.ini
```diff
@@ -1,8 +1,8 @@
 [tox]
 envlist =
-    py{36,37,38,39}-dj{22,31,32}
-    py310-dj{31,32}
-    py{38,39,310}-djmain
+    py{36,37,38,39}-dj{22,31}
+    py{36,37,38,39,310}-dj32
+    py{38,39,310}-dj{4.0,main}
```
adamchainz, line 5: I don't think this matches the below `dj40` dependency declaration with the `.`, it should be removed:

```suggestion
    py{38,39,310}-dj{40,main}
```

hramezani: Yes, you are right. it was a typo!

setup.py
hramezani: Good point. I update the `install_requires` and removed `tests_require`

hramezani:
Thanks @adamchainz for your review!

MRigal (approved):
LGTM! Good that we had no test or setup depending on pytz. Thanks @hramezani !

PR #516, opened 2021-12-26
Allow running tests using sqlite

jayvdb:
## Problem

Distro build workers do not want to install postgres in order to run the tests.

## Solution

Added capability to use sqlite instead using env flag `SQLITE`

## Commandments

- [ ] Cover it with tests.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -14,7 +14,7 @@ if os.environ.get('SQLITE'):
 else:
     DATABASES = {
         "default": {
-            "ENGINE": "django.db.backends.postgresql_psycopg2",
+            "ENGINE": "django.db.backends.postgresql",
             "NAME": os.environ.get("DB_NAME", "modelutils"),
             "USER": os.environ.get("DB_USER", 'postgres'),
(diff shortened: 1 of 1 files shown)


tests/settings.py
```diff
@@ -4,16 +4,25 @@
...
+if os.environ.get('SQLITE'):
+    DATABASES = {
+        'default': {
+            'ENGINE': 'django.db.backends.sqlite3',
+        },
+    }
+else:
+    DATABASES = {
+        "default": {
+            "ENGINE": "django.db.backends.postgresql_psycopg2",
```
auvipy, line 17: "django.db.backends.postgresql_psycopg2", should be just postgresql as per new django releases

auvipy (commented):
looks good

tests/settings.py
jayvdb: Yup, that will work dj2.2+ ; will update..

jayvdb:
ping @auvipy. I can merge but would be bad practise IMO.

foarsitter (approved):

PR #519, opened 2022-02-24
`soft` kwarg for SoftDeletableQuerySet

hamraa:
## Problem

We cannot call normal delete() on a SoftDeletableQuerySet

## Solution

Just add a `soft` kwarg for delete() method of SoftDeletableQuerySet


hramezani:
Thanks @hamraa for the patch.
I think it is `SoftDeletableQuerySetMixin` and based on the name we expect the soft delete behavior.
I am not a project lead here, But I think we can't move forward with this patch.

PR #521, opened 2022-03-18
Fix AutoLastModifiedField default cache

nesaro:
## Problem

AutoLastModifiedField default cache persists for the life of the instance of the field

## Solution

I copied a simplified version of the django implementation that returns the callable value if the default is a callable

## Commandments

- [X] Cover it with tests.
- [X] Update `CHANGES.rst` file to describe the changes, and quote according issue with `GH-<issue_number>`.
- [X] Pay attention to backward compatibility, or if it breaks it, explain why.
- [X] Update documentation (if relevant).


nesaro:
The issue seems not to happen if the model is stored in the database; it seems as if there is a second call that to default on write. I updated the test to reflect the condition that makes it fail

PR #545, opened 2022-11-15
Add support for Python 3.11 and test 3.12-dev

hugovk:
## Problem

Python 3.11 was [released on 2022-10-24](https://discuss.python.org/t/python-3-11-0-final-is-now-available/20291?u=hugovk) 🚀 

[image](https://discuss.python.org/t/python-3-11-0-final-is-now-available/20291?u=hugovk)

We should test it and declare support via the Trove classifier.

## Solution

Add to tox.ini for local and CI testing, add to GitHub Actions for CI, and add the Trove classifier to declare support.

Also bump GitHub Actions and add colour to CI logs for readability.

## Commandments

- [x] Cover it with tests.


foarsitter:
Is 3.12-dev allowed to fail? Otherwise it looks good to me. 

Needless to say it would be easier that the PR only adds the changes the title says.

hugovk:
> Is 3.12-dev allowed to fail? Otherwise it looks good to me.

No, it can also fail the build. I think that's fine: 3.12 is almost in beta and should be pretty stable.

* If there's a failure, it's good to know about it sooner rather than later, so we can fix it here and be ready for the 3.12 release, and also not to block people who depend on django-model-util for their own prep.

* If it's not a quick fix, it's easy to temporarily skip `3.12-dev` in the build.

* If it's a problem in CPython itself, even better, we can report upstream and get it fixed for everyone before the full release!

> Needless to say it would be easier that the PR only adds the changes the title says.

I've changed the title to also say we're testing `3.12-dev` here, or can also split it out to another PR if you prefer.

foarsitter:
What you describe in point three is a good one.

foarsitter (approved):

foarsitter (approved):
@hugovk forgot about this PR but found it again wondering why there was no 3.11 in the matrix. Resolved the conflict and I think we should merge It even 3.12 is failing. Agree?

hugovk:
Aha! Good to see a 3.12 failure, it's a valid failure showing there's something to fix in django-model-util (this is bullet point 1 above :)


```pytb
    File "/home/runner/work/django-model-utils/django-model-utils/model_utils/__init__.py", line 1, in <module>
      from pkg_resources import DistributionNotFound, get_distribution
  ModuleNotFoundError: No module named 'pkg_resources'
```

https://docs.python.org/3.12/whatsnew/3.12.html#removed says:

> easy_install, pkg_resources, setuptools and distutils are no longer provided by default in environments created with venv or bootstrapped with ensurepip, since they are part of the setuptools package. For projects relying on these at runtime, the setuptools project should be declared as a dependency and installed separately (typically, using pip).

I've pushed a commit to also install setuptools when we're on 3.12 or higher.

---

Alternatively, we're using pkg_resources to get the distribution version. Instead we can use `importlib.metadata` from the stdlib from 3.8+ (and a backport for 3.7), like this:

```python
try:
    # Python 3.8+
    import importlib.metadata as importlib_metadata
except ImportError:
    # Python 3.7 and lower
    import importlib_metadata
import importlib.metadata

__version__ = importlib.metadata.version(__name__)
```

Or, we're in luck, Python 3.7 is end-of-life this month (https://devguide.python.org/versions/) so we could drop that backport and simplify it to:

```python
import importlib.metadata
__version__ = importlib.metadata.version(__name__)
```

That would be my recommendation. What do you think?

foarsitter:
Thanks for providing an extensive explanation about the issue, couldn't find an easy solution myself. 

Using `importlib` is a more future proof approach I suppose so has my preferences.

Do you want to update the pull-request or merge this and use importlib when 3.7 is EOL?

foarsitter:
We are close enough if you ask me. 3.7 users won't miss any new features or something so can stick to 4.3.1.

setup.py
```diff
@@ -43,6 +43,7 @@ def long_desc(root_path):
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
```
foarsitter, line 46: Should we add 3.12 here?

.github/workflows/test.yml
```diff
@@ -2,14 +2,17 @@ name: Test
...
+
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12-dev']
```
foarsitter, line 15: Removing 3.7 here and we can merge

hugovk: Removed!

hugovk:
I need to upgrade Read the Docs to use a new Python version than 3.7. Coming up!

CHANGES.rst
```diff
@@ -16,6 +16,7 @@ To be released
 - Drop support for `Django < 3.2`
 - Drop support for `Python 3.6`
 - Confirm support for `Django 4.1`
+- Add support for `Python 3.11` (GH-#545)
```
foarsitter, line 19: Ok, last one: Drop support for `Python 3.7` is missing

foarsitter:
Since this PR contains a lot of different changes squashing doesn't seem right, thoughts?

hugovk:
Either way is fine for me, I usually don't worry too much it :)

foarsitter:
That suits me well, reading git history still isn't something I do for leisure. 

Thanks for your contribution!

blag:
Can we get a release with this change? The current available version, 4.3.1, has an issue on the latest Pythons.

PR #552, opened 2023-01-23
Add py.typed file

patrick91:
This PR adds a py.typed file, it should fix the following error when using mypy and/or pyright

```
Stub file not found for "model_utils. models" Pylance (reportMissingTypeStubs)
```

CleanShot 2023-01-23 at 22 04 50@2x

Ref: https://peps.python.org/pep-0561/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,9 @@ setup(
     zip_safe=False,
     package_data={
         'model_utils': [
-            'locale/*/LC_MESSAGES/django.po', 'locale/*/LC_MESSAGES/django.mo'
+            'locale/*/LC_MESSAGES/django.po',
+            'locale/*/LC_MESSAGES/django.mo',
+            'py.typed',
         ],
     },
 )


foarsitter (approved):

realsuayip:
@foarsitter

Does it really make sense to add `py.typed` to untyped code? Due to this, `mypy` now complains about the untyped calls to model-utils.

Previously, mypy would ignore untyped modules due to missing `py.typed`.

foarsitter:
Can you post your errors?

realsuayip:
For example, this code

```python
from model_utils import Choices

t = Choices()
```

results in

```
__init__.py:1: error: Module "model_utils" does not explicitly export attribute "Choices"  [attr-defined]
__init__.py:3: error: Call to untyped function "Choices" in typed context  [no-untyped-call]
```

the first error is probably due to missing `__all__` declaration, I'm not sure.

the second error is more annoying since it appears every time you use something from the lib, since everything is untyped.

I think the previous behavior is more desirable until the lib is properly type checked.

foarsitter:
Feel free to submit a PR to add typing. Fixing your two issues should be an easy fix I guess. 

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. [pep-0561](https://peps.python.org/pep-0561/#packaging-type-information)

Since we are not planning to add a stub package `py.typed` is the way to go.

If you are using mypy you want to know if something is untyped but if you find that behavior annoying you can ignore errors for packages: https://mypy.readthedocs.io/en/latest/config_file.html#confval-ignore_errors

realsuayip:
This bit me again, last time I had decided not to upgrade. I still think adding `py.typed` to completely untyped package is just creates inconvenience.

Anyways, this time I decided to ignore errors by adding this line to my mypy config:

`untyped_calls_exclude = "model_utils"`

foarsitter:
@realsuayip the 5.0.0 release will arrive shortly on pip and adds support for typing. 

After upgrading to 5.0.0 I would encourage you to raise an issue if you encounter typing issues.

PR #561, opened 2023-04-06
Swedish translation

tjwalch:
## Problem

Swedish translation was missing

## Solution

Provided Swedish translation


foarsitter:
@tjwalch can you rebase and add the following entry to the changelog? 

```
To be released
--------

Swedish translation (GH-#561)
```

PR #563, opened 2023-04-06
Postgresql docker container for local development

foarsitter:
## Problem

It took me some minutes to figure out what is required to run the tests locally. This PR updates the environment variables so they align with the postgres docker container and adds a docker-compose.yml to have a testing database without persisting volume.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,9 @@ name: Test
 on: [push, pull_request]
 
 env:
+  POSTGRES_USER: postgres
+  POSTGRES_PASSWORD: postgres
+  POSTGRES_DB: postgres
   FORCE_COLOR: 1
 
 jobs:
@@ -18,9 +21,9 @@ jobs:
       postgres:
         image: postgres:12
         env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
-          POSTGRES_DB: postgres
+          POSTGRES_USER: ${{ env.POSTGRES_USER }}
+          POSTGRES_PASSWORD: ${{ env.POSTGRES_PASSWORD }}
+          POSTGRES_DB: ${{ env.POSTGRES_DB }}
         ports:
         - 5432:5432
         options: >-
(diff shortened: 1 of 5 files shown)


gmcrocetti:
Hey @foarsitter , I can help you over here :). LMK in case you need a review / hand.

.github/workflows/test.yml
```diff
@@ -18,9 +21,9 @@ jobs:
       postgres:
         image: postgres:12
```
gmcrocetti, line 22: Hey @foarsitter :),
What you think about using the alpine image instead ? It will speed things up a bit

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+services:
+  postgres:
+    image: postgres:12
```
gmcrocetti, line 3: same as [here](https://github.com/jazzband/django-model-utils/pull/563/files#r1534767315) - alpine tag.

.github/workflows/test.yml
foarsitter: Never thought about it but will definitely be an improvement.

.github/workflows/test.yml
```diff
@@ -16,11 +19,11 @@ jobs:
 
     services:
       postgres:
-        image: postgres:12
+        image: postgres:13-alpine
```
gmcrocetti, line 22: Hey @foarsitter , sorry for bringing this up again. This is by no means nitpicking, just a honest question.

Is there any reason to use version 13 instead of the latest available ([16-alpine](https://hub.docker.com/layers/library/postgres/16-alpine/images/sha256-6c070fa97f2e655563c228b7ff1db4b55a1abc338c64cf739a9099812e2576a9?context=explore)) ?

foarsitter: 13 is the lowest version compatible with main: https://docs.djangoproject.com/en/dev/ref/databases/#postgresql-notes

The docs don't mention an upper limit so Im not sure if 4.2 is compatible with 16. Although our needs are very basic so probably will.

gmcrocetti: I see. I agree with your decision :)

PR #566, opened 2023-04-18
Update where `ConnectionDoesNotExist` is imported from

mthuurne:
## Problem

`ConnectionDoesNotExist` does not exist in `django.db.utils` in `django-stubs`, as it was moved to `django.utils.connection` in Django 3.2.  At some point in the future, I expect it will also only be available from its new location in Django itself.

## Solution

Import it from its new location.

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/tests/test_models/test_softdeletable_model.py
+++ b/tests/test_models/test_softdeletable_model.py
@@ -1,5 +1,5 @@
-from django.db.utils import ConnectionDoesNotExist
 from django.test import TestCase
+from django.utils.connection import ConnectionDoesNotExist
 
 from tests.models import SoftDeletable
 


CHANGES.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Not released yet
+----------------
+
+- Update where `ConnectionDoesNotExist` is imported from
```
foarsitter, line 7: Changes to the test-suite don't need appearance in the changelog

mthuurne: I've removed the changelog entry and rebased on current `master`.

foarsitter (approved):

foarsitter:
Thanks @mthuurne

PR #567, opened 2023-04-19
Remove arguments from `InheritanceQuerySetMixin._clone()`

mthuurne:
## Problem

`InheritanceQuerySetMixin._clone()` accepts and then ignores a `klass` and `setup` argument, plus any keyword arguments. The `QuerySet._clone()` method has no arguments in either Django 3.2 or 4.1 and when using a static code checker like `mypy`, this is reported as incompatible base classes for `InheritanceManager`.

## Solution

I'm assuming the arguments were necessary for a version of Django that is no longer supported by `django-model-utils`. Therefore I removed the extra arguments.

## Commandments

- [x] Cover it with tests. (The `_clone()` method was already covered.)

I chose to not add an entry in `CHANGES.rst`, as this is a cleanup that should not be noticeable to users.

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -89,7 +89,7 @@ class InheritanceQuerySetMixin:
         chained.__dict__.update(update)
         return chained
 
-    def _clone(self, klass=None, setup=False, **kwargs):
+    def _clone(self):
         qs = super()._clone()
         for name in ['subclasses', '_annotated']:
             if hasattr(self, name):


foarsitter:
This PR will break backwards compatibility right? 

Is it an idea to just accept `**kwargs` and emit a `DeprecationWarning` when `kwargs > 0`?

mthuurne:
> This PR will break backwards compatibility right?

It doesn't break any existing functionality, but instead of any passed arguments being silently ignored, a `TypeError` will be raised. So it does have the potential to escalate pre-existing problems in calling code.

However, as `_clone()` is not a public method, the only caller would be `QuerySet` or a subclass of it. So the typical application code will not be calling it directly and Django 3.2 always calls it without arguments.

> Is it an idea to just accept `**kwargs` and emit a `DeprecationWarning` when `kwargs > 0`?

Yes, but if we want to be that careful, probably it should catch both `*args` and `**kwargs`, because calling code might use positional arguments instead.

I'll leave the decision up to you: personally, I don't think it's likely this change will break any existing code, but if you prefer to be extra cautious, I'll make it accept `*args` and `**kwargs` and add the `DeprecationWarning`.

foarsitter:
Extra cautiousness isn't necessary since your argument about the private nature of the method is a good one.

foarsitter (approved):

PR #556, opened 2023-02-05
Don't use `post_init` signal for initialize tracker

meanmail:
## Problem

In one of the latest versions, sentry-sdk added logging of the launch of django signal handlers. Due to the fact that the `post_init` signal handler is used to initialize the tracker, this leads to a significant decrease in performance, and also occupies the sentry logs with useless entries, which makes it impossible to analyze performance in Sentry

image

## Solution

Don't use `post_init` signal for initialize tracker

## Commandments

- [x] Write PEP8 compliant code.
- [x] Cover it with tests.
- [x] Update `CHANGES.rst` file to describe the changes, and quote according issue with `GH-<issue_number>`.
- [x] Pay attention to backward compatibility, or if it breaks it, explain why.
- [x] Update documentation (if relevant).

PR #572, opened 2023-05-17
Connect signals to specified class only

yuekui:
## Problem

Currently the `initialize_tracker` was called for all models which is not efficient and flooded in Sentry performance logs

## Solution

Connect the signal to specified class only with least changes
Related issue: https://github.com/jazzband/django-model-utils/pull/556

## Commandments

- [ ] Cover it with tests.


foarsitter:
With https://github.com/jazzband/django-model-utils/pull/556 merged, is this PR still relevant?

PR #573, opened 2023-05-30
fix: Use proper column name instead of attname

last-partizan:
From django comment in `db/models/fields/__init__.py`:

```
# A guide to Field parameters:
#
#   * name:      The name of the field specified in the model.
#   * attname:   The attribute to use on the model object. This is the same as
#                "name", except in the case of ForeignKeys, where "_id" is
#                appended.
#   * db_column: The db_column specified in the model (or None).
#   * column:    The database column for this field. This is the same as
#                "attname", except if db_column is specified.
#
# Code that introspects values, or does other dynamic things, should use
# attname. For example, this gets the primary key value of object "obj":
#
#     getattr(obj, opts.pk.attname)
```

## Problem

Currently there is not problem, unless someone used `db_column` on their parent link field.

## Solution

This changes attribute name following django recommendations.


foarsitter:
Do you think you can provide a test-case for the scenario you are describing?

last-partizan:
Sure, updated test cases.

foarsitter:
Thanks for providing a test case. I confirmed the issue and fix locally. 

Can you update `CHANGES.rst`?

foarsitter (approved):

foarsitter:
Thanks for your time and contribution @last-partizan!

Issue #100, opened 2014-01-06
MonitorField with 'when' inits to now?


gracecodes144:
In my use-case, I have a StatusField with three options and two MonitorFields that use 'when' to watch for a specific status. See abbreviated model below.

When I initialize a new object, it appears that both active_date and deprecated_date are set to the current time. Is this the intended behavior? It would make more sense to me if these MonitorFields were left Null...

```
class Element(models.Model):
    STATUS = Choices('draft', 'active', 'deprecated')
    status = StatusField()
    created_date = models.DateTimeField('Element creation date', auto_now_add=True, null=False, editable=True)
    active_date = MonitorField('Element active date', monitor='status', when=['active'], null=True)
    deprecated_date = MonitorField('Element deprecation date', monitor='status', when=['deprecated'], null=True)
```

PR #577, opened 2023-07-19
`MonitorField` - Change default to None when the field is nullable 

gmcrocetti:
## Problem

I recently came across a non-intuitive behavior, IMHO. The MonitorField has its default value set to now even though marked as nullable. Please, I'm open to hearing the opinion case this is expected behavior.

## Solution

After catching up the discussion in https://github.com/jazzband/django-model-utils/issues/100 I decided to implement it. The discussed solution is to set its default value to None when null=True.

If **might break** backward compatibility at the edge case of someone marking the field as null=True and not expecting default to be None.

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@ Changelog
 
 To be released
 --------------
+- Add deprecation warning for MonitorField. The default value will be `None`
+  instead of `django.utils.timezone.now` - when nullable and without a default.
 - Add Brazilian Portuguese translation (GH-#578)
 - Don't use `post_init` signal for initialize tracker
 
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -1,5 +1,6 @@
 import secrets
 import uuid
+import warnings
 from collections.abc import Callable
 
 from django.conf import settings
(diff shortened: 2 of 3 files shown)


foarsitter:
We should plan a deprecation path don't you think @gmcrocetti?

gmcrocetti:
> We should plan a deprecation path don't you think @gmcrocetti?

Yes, we should. Is there any major release in the near sight ? Do you have other suggestion ?

foarsitter:
My cup of tea: A DeprecationWarning when the behavior is encountered and an environment variable to enable the new behavior so we can release 4.5, removing both when 5.0 is released?

gmcrocetti:
Humm...I initially liked the idea but I'm not so sure after getting my hands dirty. What if we instead split this new behavior in two steps:
1. Raise a deprecation warning - as you said - for `4.5.0` in this existing PR;
2. We keep the original issue and target it for `5.x.x` - I open a new PR with the actual change.

I believe it will be better cause we keep the codebase clean and also don't fall into the trap of "remembering" to remove "deadcode" added by the feature flag check.

WDYT ?

foarsitter (approved):
Very well @gmcrocetti, perfect solution!

PR #285, opened 2017-07-10
Feature/disable model save signals

ebsaral:
## Problem

When saving an object, I want to disable pre_save/post_save signals.

## Solution

I overwrote the save method of the base model class.

## Commandments

- [x] Write PEP8 compliant code.
- [x] Cover it with tests.
- [x] Update `CHANGES.rst` file to describe the changes, and quote according issue with `GH-<issue_number>`.
- [x] Pay attention to backward compatibility, or if it breaks it, explain why.
- [x] Update documentation (if relevant).

PR #582, opened 2023-10-26
Remove SaveSignalHandlingModel

adamchainz:
## Problem

`SaveSignalHandlingModel` is out of sync with upstream Django, missing bug fixes and completely broken on Django 5.0. I don’t want to check it against all supported Django versions, and I expect it will go out of sync again.

`SaveSignalHandlingModel` was added in #285 but hasn't seen any development since. It seems unpopular since there have been no issues reported, and the only results on GitHub code search are from django-model-utils copies: https://github.com/search?q=SaveSignalHandlingModel+language%3APython&type=code&l=Python

Also, the class is probably unnecessary in most cases since [`bulk_create()`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create) bypasses signals, so an easy replacement:

```python
MyModel.objects.bulk_create([MyModel(myfield="thing")])
```

Bypassing a specific receiver can be done by adding a thread-local flag to those receivers.

## Solution

Drop `SaveSignalHandlingModel`.

## Commandments

- [x] Cover it with tests.


foarsitter:
I have no objection against dropping `SaveSignalHandlingModel` but we are breaking peoples code with it. A solution that comes to my mind is keep the `SaveSignalHandlingModel` model but remove the implementation of `save_base`. This way we can add a deprecation warning in v4.4 and remove it in v5.

adamchainz:
If we remove `save_base` the class will exist but silently no longer work. I think that is unpreferable compared to a clear `ImportError` .

foarsitter (approved):

PR #590, opened 2024-03-22
Levels variable is never set

foarsitter:
Looking into the uncovered lines of code (at this moment 757 out of 768 lines are covered) I noticed that the `levels` variable in `managers.py` is never being set apart from `None`.

According to the test-suite it can be removed safely. 

Can someone review this removal?

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -44,9 +44,7 @@ class InheritanceQuerySetMixin:
         self._iterable_class = InheritanceIterable
 
     def select_subclasses(self, *subclasses):
-        levels = None
-        calculated_subclasses = self._get_subclasses_recurse(
-            self.model, levels=levels)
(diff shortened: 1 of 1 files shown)


gmcrocetti (approved):
Hi @foarsitter.
The change looks solid.
I investigated a bit about the reasoning of those changes and the majority of it was introduced by two PRs / commits:
* https://github.com/jazzband/django-model-utils/commit/93a1a44ee788467cee64bf7b48a350ad614bd9ff
* https://github.com/jazzband/django-model-utils/pull/161/files

They date back to 10 years ago and it targeted Django < 2.

Also, changes look consistent with `git grep levels`:
image
All that to say LGTM, nice job.

foarsitter:
@gmcrocetti thanks for checking out. Wasn't confided enough to merge it by my own so thanks for your approval!

PR #591, opened 2024-03-22
Explicitly re-export names from `__init__` module

mthuurne:
By listing names intended for export in the `__all__` module-level variable, static code checkers like mypy are able to verify that importing a name from a module that itself imported it is by design rather than by accident.

As flake8 understand `__all__`, this also removes the need for warning suppression comments.


foarsitter (approved):

foarsitter:
I like the gradual improvements, thanks @mthuurne

PR #510, opened 2021-12-04
Switch from freezegun to time-machine.

adamchainz:
## Problem

freezegun is slow.

## Solution

time-machine is [more accurate](https://adamj.eu/tech/2020/06/03/introducing-time-machine/) and [faster](https://adamj.eu/tech/2021/02/19/freezegun-versus-time-machine/).

## Commandments

- [x] Write PEP8 compliant code.
- [x] Cover it with tests.
- [n/a] Update `CHANGES.rst` file to describe the changes, and quote according issue with `GH-<issue_number>`.
- [n/a] Pay attention to backward compatibility, or if it breaks it, explain why.
- [n/a] Update documentation (if relevant).

PR #592, opened 2024-03-22
Revert "Switch from freezegun to time-machine."

foarsitter:
Reverts jazzband/django-model-utils#510


foarsitter:
#593 fixes it so a revert isn't needed.

PR #599, opened 2024-03-22
`MonitorField` - Change default to None when the field is nullable

gmcrocetti:
## Problem

`MonitorField`'s default value is `django.utils.timezone.now` even when `null=True`. This behavior is misleading and should be changed. Despite that, some users might rely on this very same behavior and that's why #577 introduced a deprecation warning when such behavior is encountered.

The goal of this PR is make the field use `None` as default when initialized with `null=True` and without a default - of course the provided argument for `default` takes precedence.

Original discussion in https://github.com/jazzband/django-model-utils/issues/100

## Solution

Remove the deprecation warning introduced in #577 and implement the desired behavior: use `None` when the field is **nullable** and **no** default is provided.

This change requires a major release (`5.x.x`) as discussed with @foarsitter in #577 .

## Commandments

- [x] Cover it with tests.


foarsitter:
Thanks @gmcrocetti

Issue #558, opened 2023-03-16
Add type annotations


mthuurne:
## Problem

When type-checking a model that uses `InheritanceManager` using `mypy`, the following error is displayed:

```py
app/models.py:39: error: Could not resolve manager type for
"app.models.MyModel.objects"  [django-manager-missing]
        objects = InheritanceManager()
        ^
```

According to [this Stack Overflow post](https://stackoverflow.com/questions/73165489/is-there-a-fix-for-inheritancemanager-breaking-static-type-checking), the origin of this problem is the `django-stubs` mypy plugin being unable to find `InheritanceManager` because `django-model-utils` does not provide type annotations.

According to the linked post, no actual annotations are required, just a marker that the package is annotated. But I think real annotations would be good to have anyway, both for users and developers of  `django-model-utils`.

I am working on adding annotations to `django-model-utils`. I don't know if I'll have enough time to completely annotate it, but I'll share what I have in any case, probably tomorrow.

## Environment

- Django Model Utils version: 4.3.1
- Django version: 3.2
- Python version: 3.10
- Other libraries used, if any: `django-stubs` 1.16.0

mthuurne:
I don't know yet when I'll have time to finish the annotations, but I'll submit the fixes as separate PRs, both to make sure they can be included sooner and to simplify the review of the annotations PR later.

Mostaard:
Hi! I was wondering whether you have made any progress on this? I would like to use types too. Can I help?

mthuurne:
I haven't made progress since last year. I'm still waiting for my cleanups PRs #566 and #567 to be reviewed.

foarsitter:
@mthuurne both PR's you are mentioning are merged. Are you willing to take the further lead on typing?

mthuurne:
I'll pick it up again.

I split off one cleanup that can already be applied: #591.

mthuurne:
More cleanups: #594, #595, #596.

mthuurne:
Even more cleanups: #597, #598.

I also updated [the branch](https://github.com/ProtixIT/django-model-utils/commits/type-annotations) on top of the latest `master`.

foarsitter:
Type annotations isn't a job that has to be done in one haul but you are getting pretty far. By any change, is there a way to make smaller chunks out of [the branch](https://github.com/ProtixIT/django-model-utils/commits/type-annotations)?

mthuurne:
[...]

PR #601, opened 2024-03-25
Enable static type checks using mypy in CI

mthuurne:
This is a first step in adding type annotations. See #558 for discussion.

Some things to pay attention to during review:

I'm using `from __future__ import annotations` to force postponed evaluation of annotations. This has several advantages:
- it allows modern syntax like `dict[str, str]` or `int | None` to be used in Python versions where the interpreter doesn't support that syntax yet
- it avoids most problems when using a type argument to for example `QuerySet` in Django 3.2, where `QuerySet` does not yet inherit from `typing.Generic`
- it allows using types in annotations that are defined later in the source file
- it allows using types in annotations that we can't import at runtime because it would create a circular import

I added `typing_extensions` as a runtime dependency. This makes it possible to use both runtime and static features from newer versions of Python in older versions as well. In particular, `typing.Self` simplifies a lot of annotations, but was only added to the standard library in Python 3.11.

It is possible to only use static features of `typing_extensions` and put the import within a `if TYPE_CHECKING:` guard. It makes maintenance a bit more cumbersome, but it would eliminate the runtime dependency. As a lot of other libraries are using `typing_extensions` already, I'm not sure if it avoiding the dependency in `django-model-utils` would actually make a difference in any real-world application. But if you want me to make that change, let me know.

I've used Python 3.8 in the tox configuration to run mypy, because that is the one that also runs flake8 and isort. I'm not sure that's a good reason though. Another option would be to use Python 3.12, as that is the fastest. Note that mypy's results do not depend on the Python version it runs under: it always has access to all typing features it supports.

The type checking of the unit tests requires all imported libraries to exist, which is why I added `time_machine` as a dependency of the mypy env in tox. If we want to avoid that duplication, maybe it would make sense to create a `requirements-testlib.txt` or add a `testlib` extra to `setup.py`.

It would also be possible to move `time_machine` into `requirements-test.txt`, but installing the test tools and runtime dependencies like `psycopg2-binary` isn't necessary for mypy to work and I wanted to avoid slowing down CI.

### DIFF
--- a/model_utils/__init__.py
+++ b/model_utils/__init__.py
@@ -7,6 +7,6 @@ try:
     __version__ = importlib.metadata.version('django-model-utils')
 except importlib.metadata.PackageNotFoundError:  # pragma: no cover
     # package is not installed
-    __version__ = None
+    __version__ = None  # type: ignore[assignment]
 
 __all__ = ("Choices", "FieldTracker", "ModelTracker")
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,16 @@
+[mypy]
+implicit_reexport=False
+pretty=True
+show_error_codes=True
+strict_equality=True
+warn_redundant_casts=True
+warn_unreachable=True
+warn_unused_ignores=True
+
+mypy_path = $MYPY_CONFIG_FILE_DIR
+
+plugins =
+    mypy_django_plugin.main
+
+[mypy.plugins.django-stubs]
+django_settings_module = "tests.settings"
--- a/requirements-mypy.txt
+++ b/requirements-mypy.txt
@@ -1,0 +1,2 @@
+mypy==1.9.0
+django-stubs==4.2.7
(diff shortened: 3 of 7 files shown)


mthuurne:
In the summary, I wrote:

> As a lot of other libraries are using `typing_extensions` already, I'm not sure if it avoiding the dependency in `django-model-utils` would actually make a difference in any real-world application. 

In our own Django application, it turns out only `scipy` depends on `typing_extensions`, and it has it as an optional dev dependency rather than a runtime dependency. So perhaps I overestimated how many libraries are using it.

foarsitter:
Thans for your throughout description of this PR. 

Is the `typing_extensions` thing something we can solve in another way?

mthuurne:
> Is the `typing_extensions` thing something we can solve in another way?

I used it in several places in earlier versions of the annotations, but in the latest version there are no uses left. I'll drop it from this PR and we can see whether or not to add it later if/when there is a concrete use case to discuss.

foarsitter (approved):

PR #603, opened 2024-03-26
Add type annotations

mthuurne:
This implements the bulk of the changes required for #558. The annotations are still a bit rough, but they have reached a stage where testing with third party code would be useful.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,8 @@
 [run]
 include = model_utils/*.py
+
+[report]
+exclude_also =
+    # Exclusive to mypy:
+    if TYPE_CHECKING:$
+    \.\.\.$
--- a/model_utils/choices.py
+++ b/model_utils/choices.py
@@ -1,9 +1,32 @@
 from __future__ import annotations
 
 import copy
+from typing import TYPE_CHECKING, Any, Generic, TypeVar, cast, overload
+
+T = TypeVar("T")
+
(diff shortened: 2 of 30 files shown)


foarsitter:
Sure @mthuurne, take your time :)

mthuurne:
I updated the coverage configuration to exclude lines that are only processed by mypy from the coverage measurement. However, rate limiting prevented the coverage upload from being accepted, so the shown coverage stats don't reflect the latest state.

mthuurne:
Actually, it's a bit weird that a project with as little activity as this one is running into rate limiting. Is it possible that the token we're using is old and doesn't specify that this is an open source project rather than a private evaluation?

foarsitter:
If the rate limiting is an issue we should open an dedicated issue to track it. I don't have the permissions to access the settings so I cannot be of any help. 

Final confirmation: ready for merge?

mthuurne:
> If the rate limiting is an issue we should open an dedicated issue to track it. I don't have the permissions to access the settings so I cannot be of any help.
> 
> Final confirmation: ready for merge?

Yes, it's ready.

The last change should make the coverage numbers more realistic by excluding lines that aren't expected to execute, but it doesn't change any functionality. So even if the change isn't correct, it won't negatively affect a preview release.

foarsitter:
@mthuurne thanks for all the effort you have put into this! Great job!

PR #604, opened 2024-03-26
Modernize property definitions in `SplitText`

mthuurne:
This code was using old property syntax instead of `@property` decorators.

I also removed the comments, as they don't express anything that isn't already clear from the code itself.

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -196,7 +196,6 @@ class SplitText:
         self.field_name = field_name
         self.excerpt_field_name = excerpt_field_name
 
-    # content is read/write
     @property
     def content(self):
         return self.instance.__dict__[self.field_name]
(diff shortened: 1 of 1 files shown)


foarsitter:
To be fully backwards compatible I'm against removing methods (even if they are private) for the 4.5.0 release. Do we need to mention them in the change log?

mthuurne:
> To be fully backwards compatible I'm against removing methods (even if they are private) for the 4.5.0 release. Do we need to mention them in the change log?

These methods only exist to implement the property and were introduced at the same time as the property, in commit 85157347f1dd32676322ee2dd1c4ad6e65e6eddb. There was never a reason to call the private method over using the public property.

If you're thinking of subclasses that might override the method, that doesn't actually work: the property definition captures the method from the class scope. It doesn't do a lookup on `self`, so any overriding method will not be used by the property.

PR #610, opened 2024-04-09
Remove excerpt field workaround for South FakeORM

mthuurne:
South FakeORM hasn't been supported since fda2d39e, but the workaround introduced in 9147becf evaded deletion. The `NoRendered` model was removed in 9df3df99, but with that model gone, the `no_excerpt_field` argument became obsolete as well.


foarsitter:
Can you add an entry to the change log? Something like 

> SplitField no longer accepts `no_excerpt_field` as keyword  argument

foarsitter (approved):

PR #611, opened 2024-04-11
Fix `SoftDeletableModel.delete()` forwarding positional args to superclass

mthuurne:
The `soft` argument was made keyword-only, to avoid conflicts with new positional arguments that might be added to Django. It was already conflicting with the `keep_parents` argument.

This is a backward incompatible change: code that used to pass `soft` positionally will require updating. Hopefully most code was either using the default (`True`) or was already passing `soft=False` as a keyword argument.


mthuurne:
Perhaps it would be safer to not forward any positional arguments to the superclass at all. However, that would mean deviating from the interface that Django defines.

foarsitter (approved):

PR #612, opened 2024-04-16
Auto-generate manager implementation for `CustomSoftDelete`

mthuurne:
Besides requiring less code, this also allows the django-stubs mypy plugin to automatically generate a type-annotated version of the manager.

Name the manager `available_objects` because `objects` is deprecated in `SoftDeletableModel`.


foarsitter (approved):
Only touches the tests, Thanks again!

PR #613, opened 2024-04-16
Remove `JoinManager` class

mthuurne:
The `JoinManager` class doesn't provide anything that is not already provided by `JoinQueryset.as_manager()`.

My main motivation for removing `JoinManager` is that it would be too much work to make `JoinManager` work properly with type annotations, while `as_manager()` works out of the box thanks to the django-stubs plugin.

This is a backwards incompatible change, as code using `JoinManager` or `JoinManagerMixin` would have to be updated. In most cases, I expect that update to be a single-line change; see the unit test changes for an example.

I don't know whether `JoinQueryset` is still relevant functionality. It generates SQL as a string and contains custom quote code, which always makes me worried about the potential for injection attacks. Perhaps that class could be removed as well?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ To be released
 - Make `contribute_to_class()` in `StatusField`, `MonitorField` and `SplitField`
   forward additional arguments to Django
 - `SplitField` no longer accepts `no_excerpt_field` as a keyword argument
+- Remove `JoinManager` and `JoinManagerMixin`; use ``JoinQueryset.as_manager()`` instead
 
 4.4.0 (2024-02-10)
 ------------------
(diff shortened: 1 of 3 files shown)


mthuurne:
The docs (`docs/managers.rst`) will need updating as well. But before spending more time on it, I'd like to know whether you agree with removing `JoinManager`.

foarsitter:
The `JoinManagerMixin` and `JoinManager` are there only for convenience it looks like. 

I don't like breaking code so maybe we should just ignore them for mypy and remove them in de future?

mthuurne:
> I don't like breaking code so maybe we should just ignore them for mypy and remove them in de future?

That would be fine with me.

The main thing I want to avoid is a user getting a worse type checking experience because they're not aware that there is a better supported alternative.

I'll close this PR and open a new one to deprecate `JoinManager`.

PR #614, opened 2024-04-16
Remove support for assigning `None` to a `SplitField`

mthuurne:
This behavior wasn't documented and didn't fully work: it breaks as soon as you try to save a model with a `None` value.

I think removing `None` support is better than making it work:
- to make it work properly, it would have to respect the `null` constructor argument both in the implementation and in the type checking, which is a lot of work
- [the Django docs](https://docs.djangoproject.com/en/5.0/ref/models/fields/#null) recommend against using `null=True` for text fields, so the value of the feature is questionable

As it was an undocumented feature that only partially worked, I think this change doesn't break backwards compatibility, so I didn't add it to the ChangeLog.

### DIFF
--- a/model_utils/fields.py
+++ b/model_utils/fields.py
@@ -208,9 +208,6 @@ class SplitDescriptor:
     def __get__(self, instance, owner):
         if instance is None:
             raise AttributeError('Can only be accessed via an instance.')
-        content = instance.__dict__[self.field.name]
-        if content is None:
-            return None
         return SplitText(instance, self.field.name, self.excerpt_field_name)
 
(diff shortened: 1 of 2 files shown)


foarsitter:
With 73,057 Downloads last day* you are at least breaking someones code. But since it is undocumented and not working properly I'm not against it. 

*https://pypistats.org/packages/django-model-utils

mthuurne:
Does that mean the PR is fine as it is or should I add a line to the ChangeLog?

foarsitter:
That is up to you, I don't have an opinion on this one.

PR #618, opened 2024-05-02
Remove `JoinQueryset.get_quoted_query()`

mthuurne:
Pass the parameters to the DB API instead of quoting them ourselves.


charettes:
I see that this is the only changes included in 4.5.1. Does it address any bug? I would have expected to see a regression test of some sort or at least a reference to a reported issue?

Issue #533, opened 2022-09-13
FieldTracker infinite recursion on deleting object


brianmay:
## Problem

Call `obj.delete()` results in:

```
Fatal Python error: _Py_CheckRecursiveCall: Cannot recover from stack overflow.
Python runtime state: initialized

Current thread 0x00007f955a596280 (most recent call first):
  File "/home/brian/.asdf/installs/python/3.9.7/lib/python3.9/abc.py", line 119 in __instancecheck__
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/sql/query.py", line 1443 in build_filter
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/sql/query.py", line 1562 in _add_q
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/sql/query.py", line 1532 in add_q
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/query.py", line 1445 in _filter_or_exclude_inplace
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/query.py", line 1438 in _filter_or_exclude
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/query.py", line 1420 in filter
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/manager.py", line 85 in manager_method
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/django/db/models/base.py", line 724 in refresh_from_db
  File "/home/brian/.cache/pypoetry/virtualenvs/karaage-uhi1oUwy-py3.9/lib/python3.9/site-packages/model_utils/tracker.py", line 375 in inner
[...]

PR #620, opened 2024-05-23
Fixed an infinite recursion bug when deleting models related to a tracked model with a ForeignKey

nick-prat:
When deleting a model, that has a reverse relationship to a tracked model with SET_NULL on_delete

## Problem

This issue arrises on these conditions:
1. You delete a model that is target of a foreign key, with `on_delete=SET_NULL` (I believe also with PROTECT)
2. The related model that is going to be deleted has a field tracker that tracks a foreign key field
When this happens, django loads the instance from the db with all fields deferred other than id.  In this code,
```
def current(self, fields=None):
    if fields is None:
        deferred_fields = self.deferred_fields
        if deferred_fields:
            fields = [
                field for field in self.fields
                if field not in deferred_fields
            ]
        else:
            fields = self.fields
    
    return {f: self.get_field_value(f) for f in fields}
```
we skip those deferred fields, but this is where the issue arrises, we don't check `self.field_map[field]` here, so even though the FK is deferred, we still attemp to load it, as we check for, as an example, `foo`, while the deffered field is `foo_id`.

Issue talked about here as well: https://github.com/jazzband/django-model-utils/issues/533

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -106,3 +106,4 @@
 | Őry Máté <ory.mate@cloud.bme.hu>
 | Nafees Anwar <h.nafees.anwar@gmail.com>
 | meanmail <github@meanmail.dev>
+| Nicholas Prat <nprat96@gmail.com>
--- a/model_utils/tracker.py
+++ b/model_utils/tracker.py
@@ -263,7 +263,7 @@ class FieldInstanceTracker:
             if deferred_fields:
                 fields = [
                     field for field in self.fields
-                    if field not in deferred_fields
+                    if self.field_map[field] not in deferred_fields
                 ]
             else:
                 fields = self.fields
--- a/tests/models.py
+++ b/tests/models.py
@@ -24,7 +24,7 @@ from model_utils.models import (
     TimeStampedModel,
     UUIDModel,
 )
-from model_utils.tracker import FieldTracker, ModelTracker
+from model_utils.tracker import FieldTracker, ModelTracker, FieldInstanceTracker
 from tests.fields import MutableField
 
 ModelT = TypeVar('ModelT', bound=models.Model, covariant=True)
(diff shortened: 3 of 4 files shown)


napatvc:
I also encountered issue #533 in django-model-util versions 4.5.1 and 5.0.0. I’ve tested this patch, and it resolves the issue.

@foarsitter, if everything looks good on your end, would you mind merging this PR? Thanks!

foarsitter:
Hate to ask it but @napatvc can you isolate a test-case?

napatvc:
@foarsitter, I’ve successfully created a test case here: https://github.com/jazzband/django-model-utils/commit/7e512da122c2b30903be342c44b7bd2784ce8c14

The issue is that when the test fails, it gets stuck in an infinite recursion, so we’ll need an additional tool to detect this.

I have two possible solutions:

1. Using pytest-timeout – This would stop the test if it runs too long, but the downside is that it might also interrupt other tests.

2. Using threading – This could help, but it would make the code more complex and less readable.

What are your thoughts on this approach?

foarsitter:
My first thought is to mock or override a method and raise an exception if it is called more than 3 times or so.

Should it be possible to override set_saved_fields and keep an count there?

napatvc:
@foarsitter, that’s an interesting approach! I’ve updated the code here:
https://github.com/jazzband/django-model-utils/compare/4.5.1...pentadevel:django-model-utils:4.5.1-patched
Let me know if you have any feedback!

nick-prat:
@napatvc Thanks for the support on this!

foarsitter:
@napatvc looks very good! Can you rebase your PR on this branch? Than I will merge this one and yours afterwards.

nick-prat:
@foarsitter We've merged @napatvc's branch into this PR.

foarsitter (approved):

nossila (approved):

foarsitter:
```
  mypy: commands[0]> mypy model_utils tests
  tests/models.py:285: error: Function is missing a type annotation 
  [no-untyped-def]
          def set_saved_fields(self, fields=None):
          ^
  tests/test_fields/test_field_tracker.py:582: error: Function is missing a
  return type annotation  [no-untyped-def]
          def setUp(self):
          ^
  tests/test_fields/test_field_tracker.py:582: note: Use "-> None" if function does not return a value
  tests/test_fields/test_field_tracker.py:591: error: Function is missing a
  return type annotation  [no-untyped-def]
          def test_fk_delete(self):
          ^
  tests/test_fields/test_field_tracker.py:591: note: Use "-> None" if function does not return a value
  Found 3 errors in 2 files (checked 33 source files)
```
mypy is only executed against 3.8 but it is failing with some minor things that should be easy to fix.

foarsitter:
Thanks you @nick-prat and @napatvc, a wonderful collaboration!

PR #621, opened 2024-05-26
Add ignore_empty Parameter to has_changed Method

Mng-dev-ai:
## Problem
The `has_changed` method returns True if the value changes from null to "" and vice versa, which is often not what we want

## Solution
Added `ignore_empty` parameter with default value to True to ignore if value changed from null to "" and vice versa

## Tests
I will add tests if you agree on the idea of this pr


foarsitter:
Can you describe a use-case where `None` to empty string shouldn't be an actual change? 

To me this sounds a little as https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/

Issue #541, opened 2022-10-19
delete() method don't returns number of affected rows


realnot:
## Problem

As documentation says: [https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.delete](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.delete)

> Performs an SQL delete query on all rows in the [QuerySet](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet) and returns the number of objects deleted and a dictionary with the number of deletions per object type.

Instead overridden delete() method return None

**Tips**: since you are performing an update and not a delete, please consider using bulk_update.

## Environment

- Django Model Utils version: 4.2.0
- Django version:  4.1
- Python version:  3.9
- Other libraries used, if any:

mthuurne:
I'm running into the same issue when adding type annotations: mypy warns that `SoftDeletableQuerySetMixin.delete()` isn't compatible with `QuerySet.delete()`.

What would be the correct returned value: `(0, {})` because nothing was actually deleted, or should the pretend-deleted objects be counted?

The latter might be tricky, as it would have to take into account how many objects already had `is_removed` set to `True` prior to this query.

mthuurne:
Note that `SoftDeletableModel` has the same issue as `SoftDeletableQuerySetMixin` in the case soft deletion is used.

Mogost:
I think soft deletion is supposed to mimic normal deletion. So suppose you do a soft delete of something and show the user that N records have been deleted. To him, those are the real N records that have been deleted. The fact that they are soft deleted is more your internal implementation.

mthuurne:
I wrote:

> What would be the correct returned value: `(0, {})` because nothing was actually deleted, or should the pretend-deleted objects be counted?
> 
> The latter might be tricky, as it would have to take into account how many objects already had `is_removed` set to `True` prior to this query.

Mogost wrote:

> I think soft deletion is supposed to mimic normal deletion. So suppose you do a soft delete of something and show the user that N records have been deleted. To him, those are the real N records that have been deleted. The fact that they are soft deleted is more your internal implementation.

I agree that would be the most intuitive behavior from the user's perspective.

[...]

PR #622, opened 2024-06-06
chore: Replicate Django's signature at SoftDelete queryset

gmcrocetti:
## Problem

Mimics Django's API for `SoftDeletableQuerySetMixin`.

## Solution

Django's API behaves as the [following](https://docs.djangoproject.com/en/5.0/topics/db/queries/#deleting-objects):
image

Returning a tuple of type (int, dict[str, int]) where the first element is the number of objects deleted and the second the same count but for the model itself and its relations.

The proposed solution returns the number of objects affected by the update and a made up dict where the key is model's label (no relations as we don't touch it) with aforementioned count.

## Commandments

- [x] Cover it with tests.
- [x] Update `CHANGES.rst` file to describe the changes, and quote according issue with `GH-541`.

### DIFF
--- a/model_utils/managers.py
+++ b/model_utils/managers.py
@@ -363,17 +363,17 @@ class SoftDeletableQuerySetMixin(Generic[ModelT]):
     its ``is_removed`` field to True.
     """
 
-    def delete(self) -> None:
+    def delete(self) -> tuple[int, dict[str, int]]:
         """
         Soft delete objects from queryset (set their ``is_removed``
         field to True)
         """
-        cast(QuerySet[ModelT], self).update(is_removed=True)
+        model: type[ModelT] = self.model  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


model_utils/managers.py
```diff
@@ -363,17 +363,17 @@ class SoftDeletableQuerySetMixin(Generic[ModelT]):
...
     """
 
-    def delete(self) -> None:
+    def delete(self) -> tuple[int, dict[str, int]]:
         """
         Soft delete objects from queryset (set their ``is_removed``
         field to True)
         """
-        cast(QuerySet[ModelT], self).update(is_removed=True)
+        model: type[ModelT] = self.model  # type: ignore[attr-defined]
```
foarsitter, line 371: Is the `#type ignore` part still required since @mthuurne added typing?

foarsitter: Ah, I see it isn't something you added in this PR but already is in main.

foarsitter:
Thanks @gmcrocetti !

Issue #382, opened 2019-08-05
FieldTracker for AbstractModel children is still not stable


Atorich:
## Problem

I have found that `AbstractModelTrackerTests` is not quite correct.
Since it inherited from `FieldTrackerTestCase` that just implements some extra assertions, there are no actual tests for this case.

## Solution

I have fixed `AbstractModelTrackerTests` and it seems that some tests are failed for the case with models inherited from some AbstractModel with fields defined (like AbstractUser and its children)

And also the solution for abstract models from the last release:
```
 try:
            value = self.descriptor.__get__(instance, owner)
        except AttributeError:
            value = self.descriptor
```
is still not effective.

The model in the tests defines integers attribute which is ok, but at least BooleanField and CharField won't work properly (django==2.1, model_utils==3.2.2) 

#381

jcass77:
I can confirm that models that inherit from an abstract base model (like `AbstractBaseUser`) no longer seem to update correctly since upgrading to [v3.2.0](https://github.com/jazzband/django-model-utils/releases/tag/3.2.0).

Like the failing tests demonstrate, changing one of the model's fields does not actually change the value stored in that field.

Natim:
Thank you @Atorich for finding out about this.
@jcass77 do you feel like having some bandwidth to start a PR about this?

Atorich:
#370 and #331 related

jcass77:
@Natim has indicated that this is blocked upstream in Django by [#30427](https://code.djangoproject.com/ticket/30427#ticket).

ExTexan:
Would someone update me/us on the status of this?  The last activity (other than bumping the versions) seems to be:

- it was noted that "this is blocked upstream in Django by #30427"
- Then #30427 (and #16176) were fixed.

Does that mean this issue is unblocked?  Is there a fix in the works?

PR #624, opened 2024-06-12
Remove catching of `AttributeError` in `DescriptorWrapper`

mthuurne:
## Problem

I was cleaning up the type annotations in `DescriptorWrapper.__get__()`:
```py
    def __get__(self, instance: models.Model | None, owner: type[models.Model]) -> DescriptorWrapper[T] | T:
        if instance is None:
            return self
        was_deferred = self.field_name in instance.get_deferred_fields()
        try:
            value = self.descriptor.__get__(instance, owner)
        except AttributeError:
            value = self.descriptor
        if was_deferred:
            tracker_instance = getattr(instance, self.tracker_attname)
            tracker_instance.saved_data[self.field_name] = lightweight_deepcopy(value)
        return value
```

After being puzzled by the code for some time, I realized that the `value = self.descriptor` statement was inherently incompatible with the return type of the method. Also storing the descriptor rather than a value in `tracker_instance.saved_data` seems unlikely to be correct.

## Solution

I tried to find out when this exception handler runs, but it turned out it is never executed during unit testing.

Digging deeper, I found that `AbstractModelTrackerTests` was supposed to cover this code, but that test suite was marked as `skip` because it had known test failures. Re-enabling it indeed led to 4 failing test cases, all from `ModelTrackerTests`. The reason for that was simple though: the test model `TrackedAbstract` used a `FieldTracker` rather than a `ModelTracker`. After fixing that, the test suite passed.

There used to be a [bug in Django](https://code.djangoproject.com/ticket/30427) that caused these tests to fail, according to the discussion of #370. That bug was fixed in Django 4.0.

This exception handler was introduced in #367, but subsequent discussion in #370, #381 and #382 suggests that it wasn't the proper solution. As we now have a working test and that test doesn't cause `AttributeError` to be raised, I think it is safe to remove the `try`/`except`.

## Commandments

- [x] Cover it with tests.

This change should be invisible to end users, as it removes dead code. So I think the ChangeLog and docs steps can be skipped, but if you disagree, let me know and I'll update the PR.

### DIFF
--- a/model_utils/tracker.py
+++ b/model_utils/tracker.py
@@ -50,10 +50,7 @@ class DescriptorWrapper:
         if instance is None:
             return self
         was_deferred = self.field_name in instance.get_deferred_fields()
-        try:
-            value = self.descriptor.__get__(instance, owner)
-        except AttributeError:
-            value = self.descriptor
+        value = self.descriptor.__get__(instance, owner)
         if was_deferred:
             tracker_instance = getattr(instance, self.tracker_attname)
             tracker_instance.saved_data[self.field_name] = lightweight_deepcopy(value)
--- a/tests/models.py
+++ b/tests/models.py
@@ -249,7 +249,7 @@ class TrackedAbstract(AbstractTracked):
     number = models.IntegerField()
     mutable = MutableField(default=None)
 
-    tracker = FieldTracker()
+    tracker = ModelTracker()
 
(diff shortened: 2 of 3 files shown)


foarsitter:
Underlines how valuable typing can be! Thanks for your thorough investigation.

Issue #633, opened 2024-10-19
Name conflict if tracked model has "instance" field


shemigon:
## Problem

If a tracked model has the `instance` field, creating a instance of the model throws

```
TypeError: Model.__init__() got multiple values for argument 'instance'
```

## Environment

- Django Model Utils version: 5.0.0
- Django version: any
- Python version: any
- Other libraries used, if any: none

## Code examples

```
class SomeModel(models.Model):
    name = models.CharField(max_length=100)
    instance = models.CharField(max_length=100)
```
```
model = SomeModel.objects.create(name="test", instance="12.34")
```
```
Error
Traceback (most recent call last):
  File "<projectdir>/env/lib/python3.11/site-packages/django/test/utils.py", line 446, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "<projectdir>/tests/testapp/tests/test_model.py", line 273, in test_model_with_instance_property
    model = SomeModel.objects.create(name="test", instance="test")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<projectdir>/env/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<projectdir>/env/lib/python3.11/site-packages/django/db/models/query.py", line 677, in create
    obj = self.model(**kwargs)
          ^^^^^^^^^^^^^^^^^^^^
TypeError: Model.__init__() got multiple values for argument 'instance'
```

PR #646, opened 2025-05-02
Name conflict if tracked model has field named "instance"

shemigon:
## Problem

If a tracked model has a field called `instance`, creating a instance of the model throws

```
TypeError: Model.__init__() got multiple values for argument 'instance'
```

See issue https://github.com/jazzband/django-model-utils/issues/633

## Solution

Removed `instance` argument from the new model init method and used `args[0]` instead. 

## Commandments

- [x] Cover it with tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:13-alpine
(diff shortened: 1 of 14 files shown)


livenson:
Hi, are there any plans to get this change into the upstream? We are being hit with it and wondering about prospective.

PR #664, opened 2026-09-21
Only look up deferred fields once the tracker is initialised

uptickmetachu:
Hi everyone, thanks for maintaining `django-model-utils`!... Although not sure where we are sitting with the status of jazzband.

This PR introduces an optimization that greatly reduces allocations / cpu over head for read only endpoints involving instantiating many models that use the field tracker.

## What this changes

In `DescriptorWrapper.__set__`, `was_deferred` is currently computed before checking `initialized`, but is only read inside `if initialized and was_deferred`. `initialized` is `False` for every assignment `Model.__init__` makes, because `initialize_tracker()` sets `_instance_initialized` afterwards. As a result, each of those `get_deferred_fields()` calls walks `_meta.concrete_fields` and builds a set that isn't needed at that point.

The set is largest at that point because mid-construction most fields are not in `__dict__` yet. Once the instance is built, the same call returns an empty set, which is why this primarily shows up while loading rows.

This change moves the scan behind the `hasattr` check. Because `and` short-circuits, nothing about the guarded branch changes — only whether the operands are computed eagerly.

The diff is two files: the reorder and a `CHANGES.rst` line.

## Where it came from

This came up in a CPU profile of a production Django application, where `get_deferred_fields()` called from `FieldTracker` showed up as a hot spot on list endpoints. It has been running there as a monkeypatch, and I wanted to offer the same change upstream in case it is helpful.

## Numbers

The benchmark is below rather than in the tree. It reads `DescriptorWrapper.__set__` from whatever is installed, so bundling it with the change would have it compare the change against itself — and I didn't want to add a file to the repository that you might not want to carry.

It is self-contained: `pip install django django-model-utils`, save it, and run it. It declares one model with 31 tracked fields, loads 1000 rows through `Model.from_db()` against the released `__set__`, then applies the proposed one as a monkeypatch and repeats. If the installed package already carries the change, it notes that instead of running an unnecessary comparison.

Two of the three numbers have no timing variance at all: the scan count comes from wrapping `get_deferred_fields()`, and the byte total is `sys.getsizeof()` summed over the sets built by calls whose caller is `__set__`.

Against `django-model-utils==5.0.0` from PyPI, Django 5.2.17, Python 3.13.7, load time is the minimum of 7 passes:

| | scans per row | bytes built inside `__set__` | load |
|---|---|---|---|
| original | 64 | 38.6 MiB | 81 ms |
| monkeypatched | 33 | 0.0 MiB | 42 ms |

Three consecutive runs gave 81/81/84 ms and 42/42/44 ms, with the scan and byte figures identical each time. The 31-scan difference per row is exactly the number of tracked fields.

<details>
<summary>The benchmark</summary>

```python
"""What FieldTracker costs while a model instance is being built.

`DescriptorWrapper.__set__` calls `Model.get_deferred_fields()` on every
assignment to a tracked field, but reads the answer only inside
`if initialized and was_deferred`. `initialized` is False for every assignment
`Model.__init__` makes, so each call walks `_meta.concrete_fields`, builds a
set and discards it -- at its largest, since most fields are not in `__dict__`
yet.

The first run uses django-model-utils as installed. The second applies the
proposed `__set__` as a monkeypatch and repeats it.

Self-contained: `pip install django django-model-utils`, then run this file.
"""
import sys
import time

import django
from django.conf import settings

settings.configure(INSTALLED_APPS=["django.contrib.contenttypes"], DATABASES={}, USE_TZ=True)
django.setup()

from django.db import models  # noqa: E402

from model_utils import FieldTracker  # noqa: E402
from model_utils.tracker import DescriptorWrapper  # noqa: E402

ROWS = 1000


class Job(models.Model):
    """A wide model, of the shape that makes the cost visible: the scan is
    O(concrete fields) and runs once per tracked field, so the waste is the
    product of the two."""

    class Meta:
        app_label = "contenttypes"

    reference = models.CharField(max_length=40)
    name = models.CharField(max_length=200)
    description = models.TextField()
    status = models.CharField(max_length=30)
    priority = models.IntegerField()
    created = models.DateTimeField()
    updated = models.DateTimeField()
    due = models.DateField(null=True)
    started = models.DateTimeField(null=True)
    finished = models.DateTimeField(null=True)
    is_active = models.BooleanField(default=True)
    is_billable = models.BooleanField(default=True)
    customer_id = models.IntegerField(null=True)
    site_id = models.IntegerField(null=True)
    owner_id = models.IntegerField(null=True)
    assignee_id = models.IntegerField(null=True)
    team_id = models.IntegerField(null=True)
    category_id = models.IntegerField(null=True)
    parent_id = models.IntegerField(null=True)
    estimated_minutes = models.IntegerField(null=True)
    actual_minutes = models.IntegerField(null=True)
    quoted_amount = models.DecimalField(max_digits=11, decimal_places=2, null=True)
    invoiced_amount = models.DecimalField(max_digits=11, decimal_places=2, null=True)
    purchase_order = models.CharField(max_length=60)
    internal_note = models.TextField()
    customer_note = models.TextField()
    address_line = models.CharField(max_length=200)
    postcode = models.CharField(max_length=12)
    timezone = models.CharField(max_length=60)
    source = models.CharField(max_length=30)

    tracker = FieldTracker()


def patched___set__(self, instance, value):
    """The proposal: check `_instance_initialized` before scanning."""
    sentinel = '_setting_' + self.field_name
    if (hasattr(instance, '_instance_initialized')
            and self.field_name in instance.get_deferred_fields()
            and not hasattr(instance, sentinel)):
        setattr(instance, sentinel, True)
        try:
            getattr(instance, self.field_name)
        finally:
            instance.__dict__.pop(sentinel, None)
    if hasattr(self.descriptor, '__set__'):
        self.descriptor.__set__(instance, value)
    else:
        instance.__dict__[self.field_name] = value


NAMES = [field.attname for field in Job._meta.concrete_fields]
# from_db() does not convert values, so plain ints stand in for every column.
ROWS_DATA = [tuple(range(index + 1, index + 1 + len(NAMES))) for index in range(ROWS)]


def load_all():
    for row in ROWS_DATA:
        Job.from_db(None, NAMES, row)


def measure(label):
    """Load every row once counting the sets `__set__` builds, then time it."""
    scans = wasted = 0
    original = models.Model.get_deferred_fields
    set_code = DescriptorWrapper.__set__.__code__

    def counted(instance):
        nonlocal scans, wasted
        result = original(instance)
        scans += 1
        if sys._getframe(1).f_code is set_code:
            wasted += sys.getsizeof(result)
        return result

    models.Model.get_deferred_fields = counted
    load_all()
    models.Model.get_deferred_fields = original

    best = min(_one_pass() for _ in range(7))
    print(f"{label:<16}{scans / ROWS:>11.0f}{wasted / 1024 / 1024:>11.1f} MiB{best * 1000:>9.0f} ms")
    return wasted


def _one_pass():
    start = time.perf_counter()
    load_all()
    return time.perf_counter() - start


print(f"Django {django.get_version()}, Python {sys.version.split()[0]}")
print(f"{len(NAMES)} concrete fields, {len(Job.tracker.fields)} tracked, {ROWS} rows\n")
print(f"{'':<16}{'scans/row':>11}{'scan bytes':>15}{'load':>12}")

wasted_before = measure("original")
DescriptorWrapper.__set__ = patched___set__
measure("monkeypatched")

if not wasted_before:
    print("\nThe installed django-model-utils already carries the change, "
          "so both rows describe the same code.")
```

</details>

This model holds fields and does nothing else, so the percentage is an upper bound on what a real application sees. On the application this came from, a list endpoint serialising 1000 rows of a model with 84 concrete and 30 tracked fields attributed 73.7 MB of its 193.4 MB total allocation to this one line, the largest single entry in the profile, and the request went from 1.138 s to 0.891 s wall (minimum of 6). Peak memory did not move, because the sets are freed immediately — this is allocator and CPU pressure, not retention.

## Verification

- PostgreSQL: 355 passed on `master`, 355 passed on this branch. Identical.
- SQLite: 8 failed, 347 passed both ways — the same pre-existing `JoinManager` and `iterator` tests that need PostgreSQL.
- `flake8`, `isort` and `mypy` clean.

I haven't added a test for the reorder itself, as the existing `FieldTracker` tests already cover the behaviour of the branch it guards, including deferred fields, and this change preserves that. Would you like me to add a dedicated test for this, or do you feel the existing test suite is sufficient?

## Commits

The branch has a few extra commits from adding and removing the benchmark before moving it here. Would you like me to squash the branch down to a single commit for review, or do you prefer keeping the separate commits?


uptickmetachu:
Before merging I will remove the benchmark.

The benchmark is only here for ease of reproduction.

uptickmetachu:
The benchmark behind the numbers in the description. Save it anywhere and run it — it configures Django itself and needs no database, so `pip install django django-model-utils` is the only setup.

The first run measures `django-model-utils` as installed, then the proposed `__set__` goes on as a monkeypatch and the same rows are loaded again. If the installed package already carries the change, it says so rather than printing two identical rows.

```python
"""What FieldTracker costs while a model instance is being built.

`DescriptorWrapper.__set__` calls `Model.get_deferred_fields()` on every
assignment to a tracked field, but reads the answer only inside
`if initialized and was_deferred`. `initialized` is False for every assignment
`Model.__init__` makes, so each call walks `_meta.concrete_fields`, builds a
set and discards it -- at its largest, since most fields are not in `__dict__`
yet.

The first run uses django-model-utils as installed. The second applies the
proposed `__set__` as a monkeypatch and repeats it.

Self-contained: `pip install django django-model-utils`, then run this file.
"""
import sys
import time

import django
from django.conf import settings

settings.configure(INSTALLED_APPS=["django.contrib.contenttypes"], DATABASES={}, USE_TZ=True)
django.setup()

from django.db import models  # noqa: E402

from model_utils import FieldTracker  # noqa: E402
from model_utils.tracker import DescriptorWrapper  # noqa: E402

ROWS = 1000


class Job(models.Model):
    """A wide model, of the shape that makes the cost visible: the scan is
    O(concrete fields) and runs once per tracked field, so the waste is the
    product of the two."""

    class Meta:
        app_label = "contenttypes"

    reference = models.CharField(max_length=40)
    name = models.CharField(max_length=200)
    description = models.TextField()
    status = models.CharField(max_length=30)
    priority = models.IntegerField()
    created = models.DateTimeField()
    updated = models.DateTimeField()
    due = models.DateField(null=True)
    started = models.DateTimeField(null=True)
    finished = models.DateTimeField(null=True)
    is_active = models.BooleanField(default=True)
    is_billable = models.BooleanField(default=True)
    customer_id = models.IntegerField(null=True)
    site_id = models.IntegerField(null=True)
    owner_id = models.IntegerField(null=True)
    assignee_id = models.IntegerField(null=True)
    team_id = models.IntegerField(null=True)
    category_id = models.IntegerField(null=True)
    parent_id = models.IntegerField(null=True)
    estimated_minutes = models.IntegerField(null=True)
    actual_minutes = models.IntegerField(null=True)
    quoted_amount = models.DecimalField(max_digits=11, decimal_places=2, null=True)
    invoiced_amount = models.DecimalField(max_digits=11, decimal_places=2, null=True)
    purchase_order = models.CharField(max_length=60)
    internal_note = models.TextField()
    customer_note = models.TextField()
    address_line = models.CharField(max_length=200)
    postcode = models.CharField(max_length=12)
    timezone = models.CharField(max_length=60)
    source = models.CharField(max_length=30)

    tracker = FieldTracker()


def patched___set__(self, instance, value):
    """The proposal: check `_instance_initialized` before scanning."""
    sentinel = '_setting_' + self.field_name
    if (hasattr(instance, '_instance_initialized')
            and self.field_name in instance.get_deferred_fields()
            and not hasattr(instance, sentinel)):
        setattr(instance, sentinel, True)
        try:
            getattr(instance, self.field_name)
        finally:
            instance.__dict__.pop(sentinel, None)
    if hasattr(self.descriptor, '__set__'):
        self.descriptor.__set__(instance, value)
    else:
        instance.__dict__[self.field_name] = value


NAMES = [field.attname for field in Job._meta.concrete_fields]
# from_db() does not convert values, so plain ints stand in for every column.
ROWS_DATA = [tuple(range(index + 1, index + 1 + len(NAMES))) for index in range(ROWS)]


def load_all():
    for row in ROWS_DATA:
        Job.from_db(None, NAMES, row)


def measure(label):
    """Load every row once counting the sets `__set__` builds, then time it."""
    scans = wasted = 0
    original = models.Model.get_deferred_fields
    set_code = DescriptorWrapper.__set__.__code__

    def counted(instance):
        nonlocal scans, wasted
        result = original(instance)
        scans += 1
        if sys._getframe(1).f_code is set_code:
            wasted += sys.getsizeof(result)
        return result

    models.Model.get_deferred_fields = counted
    load_all()
    models.Model.get_deferred_fields = original

    best = min(_one_pass() for _ in range(7))
    print(f"{label:<16}{scans / ROWS:>11.0f}{wasted / 1024 / 1024:>11.1f} MiB{best * 1000:>9.0f} ms")
    return wasted


def _one_pass():
    start = time.perf_counter()
    load_all()
    return time.perf_counter() - start


print(f"Django {django.get_version()}, Python {sys.version.split()[0]}")
print(f"{len(NAMES)} concrete fields, {len(Job.tracker.fields)} tracked, {ROWS} rows\n")
print(f"{'':<16}{'scans/row':>11}{'scan bytes':>15}{'load':>12}")

wasted_before = measure("original")
DescriptorWrapper.__set__ = patched___set__
measure("monkeypatched")

if not wasted_before:
    print("\nThe installed django-model-utils already carries the change, "
          "so both rows describe the same code.")
```

rtpg (approved):
Disclaimer: I work with @uptickmetachu but did not work on the patch together with them

This looks right. The core thing here is getting short circuiting, and this feels like an "obvious" win here

